PR #1, opened 2017-05-13
Electron "Hello World" with Typescript and electron-prebuilt-compile

charnpreetsingh:
Added in various repo files.
To run: `npm start` ---> this will compile index.ts automatically thanks to `electron-prebuilt-compile`


ellisonbg:
This is the basic template we will need for just about anything. Merging...

PR #6, opened 2017-05-24
Basic build using jupyterlab packages

ellisonbg:
This almost works, working on debugging the last few build issues.


ellisonbg:
This is now working! Lots more to do, but I am going to merge so @charnpreetsingh can iterate. Here is a screenshot of JupyterLab running in Electron, with a fully self contained build of the frontend based on the jupyterlab npm packages:

screen shot 2017-05-24 at 3 50 44 pm

PR #7, opened 2017-05-25
Build before starting.

jdetle:
I think it makes sense to build before starting. Some more work incoming.


ellisonbg:
Thanks merging!

PR #9, opened 2017-05-26
differentiate starting and building

charnpreetsingh:
Follow up to #7 

I agree that we should have a command that lets you build all the source files and start the electron process but I also like to be able to just start the electron process with a separate command without running webpack. 

Now if you run `npm run build:all` it will compile all the source and  start the electron process.


ddavidebor:
This does not conflict with the most common electron syntaxes to release electron apps so that's ok for me :-)

ellisonbg:
Looks good, merging...

Issue #8, opened 2017-05-26
Unable to Close electron app


charnpreetsingh:
Currently you need to force quit or kill the electron process in order to close the app because JupyterLab is prompting the user if they really want to leave the page and Electron is not showing that prompt. 

We want to show our own prompt in the `onbeforeunload` event of the Electron `BrowserWindow`
https://github.com/electron/electron/blob/master/docs/api/browser-window.md#event-close

PR #10, opened 2017-05-26
remove application-extension dependenices

charnpreetsingh:
Partially addresses #8 

Removes the extensions so that  JupyterLab's `onbeforeunload` does not interfere with electron's


ellisonbg:
We will need to replace that plugin with another that has the rest of the functionality though. Merging for now...

PR #19, opened 2017-06-07
Update types and put source in one directory

charnpreetsingh:
- removed old typings and uses updated typings that are already bundled with electron (as of `1.6.10`)
- all source files are in the `src/` directory


ellisonbg:
Looks like this needs to be rebased, since I merged the other PR...

charnpreetsingh:
@ellisonbg rebased

PR #26, opened 2017-06-22
Start jupyter server in child process and launch render process with token.

lucbouchard1:
This commit fixes issue #12 . The server is launched in a child process and the token is collect using `jupyter notebook list`. The token is then injected into the frontend html using Handlebars.


ddavidebor:
WOOOOOAH

ddavidebor:
@lucbouchard1 why did you remove electron-prebuilt-compile and used vanilla typescript instead?

lucbouchard1:
@ddavidebor I was having problems with the build and changed it out of frustration. I think I have more of a grip on how to configure it now so I will change it back. I agree it's better with electron-prebuilt-compile. I'll open an issue.

ddavidebor:
For future reference: the issue is https://github.com/jupyterlab/jupyterlab_app/issues/28

PR #33, opened 2017-06-29
Bump to newest release of JupyterLab. Add splash screen.

lucbouchard1:
Updating to the newest release of JupyterLab allowed us to remove templating. The allowed adding a loading screen that makes startup feel really fast and smooth.


willingc:
Splash screen is cool!

PR #34, opened 2017-06-30
Update splash to repeat and fade at end of an iteration

nproctor:
The splash screen fade was previously hardcoded into the CSS. This fades the splash screen in "index.ts" once the app is loaded and the animation reaches the end of an iteration. The CSS was adjusted to add a delay at the beginning of an iteration.

### DIFF
--- a/src/browser/css/main.css
+++ b/src/browser/css/main.css
@@ -26,7 +26,6 @@ table { border-collapse: collapse; border-spacing: 0; }
 
 
 #universe {
-    animation: fade .4s linear 2.5s forwards;
     z-index: 1;
     position: absolute;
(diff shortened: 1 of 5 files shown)


willingc:
I ran this locally last night. It works. Love the fade. I wasn't sure what you meant by repeat in the PR title; I only saw one rotation. @nproctor The code LGTM. Nicely done.

nproctor:
The animation only repeats if the application isn't ready by the end of an iteration. So if the lab has loaded by the time the first iteration completes, you'll only see one rotation.

willingc:
Thanks for the explanation @nproctor. That makes complete sense. Looking good :sunny:

PR #42, opened 2017-07-01
Move JupyterLab menu bar to the native menu bar

### LINKED ISSUE #40: Utilize native menu bar
The JupyterLab electron app will place the menu bar content in the native menu bar instead of at the top of the window. This will be a more efficient use of screen space.

lucbouchard1:
The PR removes moves the menu bar to the top of the screen, utilizing the native menu bar. It's only been tested on my Ubuntu machine, so features should be verified on other platforms before merging. This fixes #40


src/main/menu.ts
```diff
@@ -0,0 +1,91 @@
...
+     */
+    private registerListeners(): void {
+        /* Register MENU_ADD event */
+        ipcMain.on(JupyterMenuChannels.MENU_ADD, (event: any, menu: ItemOptions) => {
+            this.addMenu(event, menu);
+        });
+    }
+
+    /**
+     * Add a menu to the menubar. Sets up click handlers
```
willingc: Perhaps:
```
* Add a menu to the menubar.
* Sets up click handlers on submenu tree items.
* Chooses menubar position of menu based on the 'id' field string.
* Lower...
```

src/ipc.ts
```diff
@@ -0,0 +1,36 @@
...
+namespace JupyterMenuChannels {
+
+    /**
+     * IPC channel for appending menus to the native menu bar
+     */
+    export
+    let MENU_ADD = 'menu-add';
+
+    /**
+     * IPC channel for recieveing click events
```
lucbouchard1: Thanks for the feedback @willingc! I will be sure to put more care into my documentation in the future. 

Looks like there might be a problem with how this is running on your computer. I don't see the 'File' menu appearing, and the upper left corner says 'Electron' instead of 'JupyterLab'. Is the 'Electron' menu item clickable? Perhaps showing the contents of file menu?

lucbouchard1:
Thanks for the feedback @willingc! I will be sure to put more care into my documentation in the future. 

Looks like there might be a problem with how this is running on your computer. I don't see the 'File' menu appearing, and the upper left corner says 'Electron' instead of 'JupyterLab'. Is the 'Electron' menu item clickable? Perhaps showing the contents of the file menu?

willingc:
@lucbouchard1 I actually think your comments are very good. My suggestions are more to get you in the habit of thinking about a future developer that doesn't know the code base. Just keep up doing what you are doing :-)

Looks like the Electron menu is showing the File menu contents. I'll be off for a few days and then back on the 5th. Have a good weekend too.

screen shot 2017-06-30 at 11 18 02 pm

src/browser/electron-extension/index.ts
```diff
@@ -9,11 +9,39 @@ import {
...
+  /**
+   * Construct a new JupyterLab object.
+   */
+  constructor(options: JupyterLab.IOptions = {}) {
+    super();
+
+    /* Remove uneeded panels */
+    each(this.shell.layout.iter(), (widget: Widget) => {
+        if (widget.id == 'jp-top-panel') {
+            this.shell.layout.removeWidget(widget);
```
ellisonbg: The correct way to remove a widget is to set its parent to null

nproctor:
The menu items are duplicating on Mac when the application is "activated".  

screen shot 2017-07-03 at 2 21 11 pm

lucbouchard1:
@ellisonbg Can you verify that I'm removing the widget properly?

ellisonbg:
Looks good, I have approved, so feel free to merge!

PR #45, opened 2017-07-07
Store window size at app close so app can be reopened in same state

lucbouchard1:
This PR adds code to store the position/dimensions of the main window when the user quits the app. This allows recreating the window that was open when the user last quit the app.

To do this, I added some generic code that allows easily storing application state (src/main/state.ts). This code will be useful as we add more features that depend on persistent state.

### DIFF
--- a/src/browser/utils-extension/nativemenu.ts
+++ b/src/browser/utils-extension/nativemenu.ts
@@ -77,7 +77,7 @@ class NativeMenu extends MenuBar implements IMainMenu {
                 nItems[i].type = (items[i].type as 'normal' | 'submenu' | 'separator');
             nItems[i].label = items[i].label;
             nItems[i].command = items[i].command;
-            console.log(items[i].args);
+
             nItems[i].args = items[i].args;
 
             if (items[i].submenu !== null)
--- a/src/main/app.ts
+++ b/src/main/app.ts
@@ -96,6 +96,7 @@ export class JupyterApplication {
     constructor() {
         this.registerListeners();
         this.server = new JupyterServer();
+        this.mainWindow = new JupyterLabWindow();
         this.menu = new JupyterMainMenu();
     }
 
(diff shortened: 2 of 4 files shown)


ddavidebor:
In this PR you're not using the JupyterLab APIs to store the window position and dimention, right ?

lucbouchard1:
@ddavidebor Correct. If we have to wait for the Jupyter server to startup before we can get our window dimensions, it's going to make the render time unbearably slow.

ddavidebor:
Isn't it a simpler solution to set a default window size that looks good with the splash screen, then once jupyter is loaded get the window size from the config and resize the window with the electron [APIs](https://github.com/electron/electron/blob/master/docs/api/browser-window.md) just before disabling the splash screen?

We shouldn't duplicate low-level features between jupyter and jupyterlab_app.

ddavidebor:
Also, the directories where the code resides are not writeable once the application is packaged.

ddavidebor:
That said, sooner or later we'll have to store something somewhere that is not a Jupiter process, such as the addresses of remove jupyter servers... :smile:

lucbouchard1:
We'll have to sit down and talk with @ellisonbg and the designers about resizing after the server has started. I really don't think that makes sense because the application doesn't only store the size of the last window, it also stores the position. That means that as soon as the server starts, we would either have to move and resize the window or close the splash window and open a separate window, both of which result in a more jagged transition from splash screen to app. Cameron also did a great job of centering the splash screen logo in any window via CSS, so I don't think that will be an issue. 

The data is being stored in the system-specific application data directory. That directory is found using the electron API. We will not be writing to the directory where the code resides. 

You're right, we will need to store some application state eventually anyway, so I think it makes sense to leverage it here.

ddavidebor:
Yeah i agree also the windows size and position has more to do with the user's PC than the notebook server, especially considering remove jupyter servers

nproctor:
On MacOS, the window size persists when you close and open a new app, but the window doesn't open at all on activate.

nproctor:
The application seems slow to activate. See issue #53.

Issue #41, opened 2017-07-01
Add debug mode


lucbouchard1:
In replacing the default electron menu bar with the JupyterLab menu bar (#40), I have taken away the ability to open the chrome developer tools. This is fine for the final product, but for development we should be able to start the app in debug mode using a command line argument. Debug mode should add an additional menu to the menu bar that enables the developer tools.

ddavidebor:
We often wack this everywhere https://github.com/sindresorhus/electron-debug

Davide

PR #46, opened 2017-07-07
Packaging on linux

ddavidebor:
This PR introduces a way to easily build the application for a Linux environment. Mac and Windows will come as soon as I have more free time (or others wanna help out). 

I still need to test if the icons and menu shortcuts work well on a fresh distro install, the workstation I work with is fairly messed up in this regard.

But it's a decent start. Mac release coming soon.


lucbouchard1:
@ddavidebor Thanks so much for doing this! Packaging worked great on my Ubuntu machine. I am running into a problem when I execute the app though. As soon as the app window opens, it just sits on the splash screen and doesn't actually start JupyterLab. This suggests to me that the Jupyter server isn't starting up. Any ideas on how to debug this?

I also have a newbie question for you. Is the use of docker just to make the packaging code platform independent, or is there another reason why it's necessary?

ddavidebor:
Docker it's not particularly necessary, you can install all the required tools to do the build, but it's convenient and building releases from a standard environment is good practice. It also has [Wine](https://www.winehq.org/) and [mono](http://www.mono-project.com/) included to build for windows.

I am running into a few issues too. For example it finds only my system default python, and not anaconda. I think the problem is that it's being launched from the desktop environment instead of from the terminal and it does not inherit all the environmental variables from the .bashrc (or similar) file. 

It should be possible to launch the development tool with the [shortcut](https://github.com/sindresorhus/electron-debug) from #41 #44

ddavidebor:
Yeah i think that's it: when launched from the terminal with `jupyterlab_app` works like a charm

ddavidebor:
mmh perhaps we should let the user search and select the correct jupyter notebook server ? sooner or later we're going to need a well-generalized UI to do that... but it's not immediate to implement.

lucbouchard1:
I agree, I created an issue (#48) to talk about it.

ddavidebor:
@lucbouchard1 Everything seems ready to build for linux.
Has anyone tested the rpm?

lucbouchard1:
@ddavidebor I haven't tested the rpm yet. I'll spin up a Fedora VM and try it out.

ddavidebor:
https://transfer.sh/RtFNo/jupyterlab_app_0.0.1_amd64.deb
https://transfer.sh/8ON6h/jupyterlab_app-0.0.1.tar.gz
https://transfer.sh/64Hs4/jupyterlab_app-0.0.1.x86_64.rpm

lucbouchard1:
I needed to yum install libXScrnSaver in order to get the application to launch in Fedora, but other than that the install worked well. There appears to be a couple bugs that are internal to the JupyterLab app, but those are separate from this. I'm going to go ahead and merge this PR.

ddavidebor:
@lucbouchard1 did you have to install it manually or was the dependency resolved automatically?

lucbouchard1:
@ddavidebor I needed to install manually. Looks like a mirror of atom/atom#13176

ddavidebor:
yep. #58 . I'll fix it in the following days.

probonopd:
Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions

[Here is an overview](https://github.com/probonopd/AppImageKit/wiki/AppImages) of projects that are already distributing upstream-provided, official AppImages.

__electron-builder__ can produce AppImages automatically for you.

ddavidebor:
I agree i'll open an issue

PR #62, opened 2017-07-14
Windows compatibility

### LINKED ISSUE #54: Make sure path separators will work on Windows
I saw some `/` in paths...

nproctor:
Fixes #54. Checks if the platform is Windows. If it is, the server is spawned with Windows compatible commands, and in app terminals are disabled.


nproctor:
There's still an issue with child processes not terminating. I'm working on it, but since #66 depends on this merge, I'll create a new issue for this.

ddavidebor:
As you prefer, I'll happily put #66 on halt and fix conflicts later

PR #80, opened 2017-07-27
Add keyboard shortcuts without adding to the menubar

nproctor:
This removes the need for an "Edit" menu by implementing keyboard shortcuts without putting them in the menu. Electron requires local shortcuts to be in the menubar, so instead the keyboard shortcuts are implemented as global shortcuts that are registered on application focus and unregistered when the application loses focus. Right now, it registers copy, cut, paste, zoom-in, and zoom-out, and others can added if needed.

### DIFF
--- a/src/browser/extensions/utils-extension/nativemenu.ts
+++ b/src/browser/extensions/utils-extension/nativemenu.ts
@@ -65,13 +65,15 @@ class NativeMenu extends MenuBar implements IMainMenu {
     private buildNativeMenu(menu: Menu): JupyterMenuItemOptions[] {
         let items = menu.items;
         let nItems = new Array<JupyterMenuItemOptions>(items.length);
-        for (let i = 0, n = nItems.length; i < n; i++) {
+        for (let i = 0; i < items.length; i++) {
             nItems[i] = {command: null, type: null, label: null, submenu: null, accelerator: null};
 
-            if (items[i].type == 'command')
+            if (items[i].type == 'command') {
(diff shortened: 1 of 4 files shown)


ddavidebor:
Why removing the edit menu?

Paste and match style is not discoverable by the user. It would also be a great place to add search and replace functionality. 
It's also a bit of a cornerstone in sw, dumb users could get lost.

lucbouchard1:
@ddavidebor JupyterLab devs decided against an edit menu in early design. You will notice that JupyterLab in the browser does not have an edit menu, @ellisonbg wanted us to be consistent. Copy and paste key bindings are pretty standard, so I don't think it's going to be particularly cumbersome for inexperienced users.

In addition, for design reasons we want to migrate to a frameless window. Since Windows puts the native menu bars in the application window instead of at the top of the screen, using a frameless window causes the native menu bar to be invisible.  Therefore on Windows, we will be including the PhosphorJS DOM-based menu bar. 
The only problem with putting the menu bar in the DOM is that as soon as you click a menu item, it causes any text area that's currently in focus to lose focus. This means executing the browser's copy/paste commands from a DOM edit menu won't actually work, because text area's can't be in focus. In other words, the edit menu is rendered useless. There might be a way to hack around this, but we decided it makes more sense to just omit the edit menu.

I know the Jupyter Notebook already implements find/replace via the command palette, and I assume that functionality is going to be supported by JupyterLab.

ddavidebor:
In this case seems like a nice idea :smile:

PR #84, opened 2017-07-28
.svg file for icon 

hzarea:
@ddavidebor I've included an .svg version of the icon. SVGs can be opened and edited in Adobe Illustrator!


ddavidebor:
Great thanks!

PR #86, opened 2017-07-29
Build and release for mac on Travis ( Fix #81 )

### LINKED ISSUE #81: Build for Mac on travis
I'll fix this, just for tracking

ddavidebor:
This PR sets up Travis to build in parallel for linux and macOS. Unfortunately, due to a [bug](https://github.com/electron-userland/electron-builder/issues/1871), i was not able to set up the (pretty great) release feature of electron-builder and had to disable it and use the one in Travis-CI. Didn't want to lock the version to an older one.

In the future, electron-builder will allow to run parallel builds and then merge all artifacts in a single, well configured release.

Thus, for now, all artifacts that will be released are built on macOS. This will allow to release the v0.1.0 and v2.0.0 from a controlled enviroment, even if the person releasing is on linux/win (mac artifacts can only be built on macs.

Someone with write access will need to change the github release secure key from the .travis.yml file following this instructions https://docs.travis-ci.com/user/deployment/releases

Tagged commits (for example from running `npm version patch/minor/major` are automatically built and deployed as github release drafts.


ddavidebor:
@lucbouchard1

ddavidebor:
I nearly got it working with the releasing using electron-builder, which is way more practical and allows for quicker builds https://github.com/fermiumlabs/jupyterlab_app/tree/electron-builder-release

lucbouchard1:
@ddavidebor Great! Can you just submit another PR when you get that working?

ddavidebor:
@lucbouchard1 sure, i've got it working but it will fail in any other repo but jupyterlab/jupyterlab_app so i'm trying to fix this last detail

ddavidebor:
#87

PR #121, opened 2017-08-18
Update JupyterLab to v0.26. Add support for native settings backend.

lucbouchard1:
This PR fixes issue #120. It updates the JupyterLab frontend to the most recent release, in addition to adding support for backend settings. The implementation works as follows:

- At build time, we run A script to extract the default settings. These will be included in the application bundle.
- When the app is launched, we check two locations for settings. In order of precedence, they are: the JupyterLab Native application data directory and the application bundle with the default settings.
- These settings are made available to JupyterLab by replacing the `IDataConnector` interface associated with the settings plugin.
- At application close, we always save the settings to the JupyterLab Native application data directory. Subsequent launches will always pick up settings from that location.


lucbouchard1:
It's worth noting that the extraction script will need to be updated for the JupyterLab v0.27 release. Schemas are stored differently in JupyterLab HEAD than they are in v0.26.

src/main/sessions.ts
```diff
@@ -213,7 +247,7 @@ class JupyterLabSessions extends EventEmitter implements ISessions, IStatefulSer
             
         });
 
-        ipcMain.once(AppIPC.LAB_READY, () => {
+        ipcMain.once('lab-ready', () => {
```
declanvk: Would it be better to replace this hard coded string with a constant, similar to what previously existed?

src/main/app.ts
```diff
@@ -93,7 +93,7 @@ class JupyterApplication implements IApplication {
      */
     constructor() {
         this._registerListeners();
-        
+
```
declanvk: Good work!

lucbouchard1: Thanks

PR #144, opened 2018-02-21
Implement disabled/deferred/ignored extensions and fix linter problems

Wietsedv:
The list of core extensions is still hardcoded in the code, but fully dynamic module loading is not possible. The solution of the main JupyterLab package is automatically rewriting the startup script to import the relevant modules before webpack does its work. This solution will not do here since the relevant file is written in typescript which means that the file cannot contain template information. Typescript does not like that.

This will probably change anyway before JLab 1.0, so I guess we do not have to worry about it too much.

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,5 +188,44 @@
     "which": "^1.3.0",
     "winreg": "^1.2.4",
     "xterm": "^3.0.2"
+  },
+  "jupyterlab": {
+    "name": "JupyterLab",
+    "version": "0.31.5",
+    "extensions": [
+      "./electron-extension",
+      "./utils-extension",
+      "@jupyterlab/codemirror-extension",
+      "@jupyterlab/completer-extension",
+      "@jupyterlab/console-extension",
+      "@jupyterlab/csvviewer-extension",
+      "@jupyterlab/docmanager-extension",
+      "@jupyterlab/faq-extension",
+      "@jupyterlab/filebrowser-extension",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -188,5 +188,44 @@
     "which": "^1.3.0",
     "winreg": "^1.2.4",
     "xterm": "^3.0.2"
+  },
+  "jupyterlab": {
```
declanvk: Do you end up using this data embedded in the `package.json` file as part of the extension loading system?

Wietsedv: This data is used to iterate over when it is determined which extensions have which status. This is copied from the main JupyterLab repo. I wanted to use this list to import the extensions, but this kind of dynamic importing does not appear to be possible. Anyway, this is redundant right now since the extensions are also listed in another file. I'll remove it.

src/browser/index.tsx
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import './require';
```
declanvk: I'm assuming that removing this is covered by the updated rules in the webpack configuration?

declanvk: Other than this, everything looks great and I'm looking forward to merging this.

Wietsedv: Thanks. That is true. This weird hook is unnecessary right now. I only forgot to commit the removal of the actual file. Now it is removed.

Issue #110, opened 2017-08-07
Expand main process debug output


lucbouchard1:
We should add a command line flag that enables a more verbose debug output. The expanded debug output should include things like the output of the Jupyter server.

declanvk:
Closed due to #165

PR #165, opened 2018-05-08
Expand debug output

declanvk:
This PR expands the amount of debug output displayed to the console by routing the render console output to the terminal.

It also adds CLI options that specify the desired level of output using the `-v` option flag. Multiple occurrences of the flag lead to more debug output.

Resolves issue raised in #110.


declanvk:
Additional feature could be differentiating main and render process debug output, waiting on megahertz/electron-log#75.

PR #219, opened 2021-09-18
Release workflow

mbektas:
Updates to GitHub Actions
- Checks for version match in the App and bundled JupyterLab packages
- Uploads installer binaries to a Release if the tag match is found.
  - Release needs to have tag as `v<AppVersion>`
  - Release needs to be of type `draft` or `pre-release`
- Release documentation is updated with new instructions

For a JupyterLab version, now we can release multiple versions of JupyterLab App.

Example Release with installer assets attached [releases/tag/v3.1.12-1](https://github.com/jupyterlab/jupyterlab_app/releases/tag/v3.1.12-1)

Fixes https://github.com/jupyterlab/jupyterlab_app/issues/205

### DIFF
--- a/Release.md
+++ b/Release.md
@@ -42,8 +42,8 @@ In order to change the JupyterLab version bundled with the App:
 
     tbump will list changes to be applied, confirm the changes to proceed with apply.
 
-2. Create a commit with the version changes and create a PR. The PR must be created from main repo and not from a fork. This is necessary for GitHub Actions to be able to attach installers to the release.
+3. Create a commit with the version changes and create a PR. The PR must be created from main repo and not from a fork. This is necessary for GitHub Actions to be able to attach installers to the release.
 
-3. GitHub Actions will automatically create installers for each platform (Linux, macOS, Windows) and upload them as release assets. Assets will be uploaded only if a release of type `draft` or `pre-release` with tag matching the JupyterLab App's version with a `v` prefix is found. For example, if the JupyterLab App version in the PR is `3.1.12-2`, the installers will be uploaded to a release that is flagged as `draft` or `pre-release` and has a tag `v3.1.12-2`. New commits to PR will overwrite the installer assets of the release.
+4. GitHub Actions will automatically create installers for each platform (Linux, macOS, Windows) and upload them as release assets. Assets will be uploaded only if a release of type `draft` or `pre-release` with tag matching the JupyterLab App's version with a `v` prefix is found. For example, if the JupyterLab App version in the PR is `3.1.12-2`, the installers will be uploaded to a release that is flagged as `draft` or `pre-release` and has a tag `v3.1.12-2`. New commits to PR will overwrite the installer assets of the release.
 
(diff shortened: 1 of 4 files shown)


Release.md
```diff
@@ -1,23 +1,49 @@
...
+    ```
+
+    Example: changing JupyterLab version (to `3.1.13`), bump from `3.1.12-3` to `3.1.13-1`
+    ```bash
+    tbump --only-patch 3.1.13-1
+    ```
+
+    tbump will list changes to be applied, confirm the changes to proceed with apply.
+
+2. Create a commit with the version changes and create a PR. The PR must be created from main repo and not from a fork. This is necessary for GitHub Actions to be able to attach installers to the release.
```
blink1073, line 45: Repeated bullet number?

mbektas: yes, thank you, good catch!

Release.md
```diff
@@ -1,23 +1,49 @@
...
-# Dependencies
 
-To build for Linux and Windows the suggested way is to use the `dockerdist:linux` and `dockerdist:windows` command that use a well-tested and fully equipped docker environment to build for both.
-To build the dmg container for macOS, a macOS machine is required. It's suggested to use Travis-CI in this regard.
-This [documentation](https://github.com/electron-userland/electron-builder/wiki/Multi-Platform-Build) explains the needed dependencies to build locally.
+## Updating the bundled JupyterLab
+
+In order to change the JupyterLab version bundled with the App:
+
+1. Update all the dependencies in [package.json](package.json) (packages with `@jupyterlab` scope and also `@jupyter-widgets/jupyterlab-manager` for ipywidgets). Versions of `@jupyterlab` packages can be found in the commits for the release tag at https://github.com/jupyterlab/jupyterlab/releases
```
blink1073, line 23: There is a federated extension for ipywidgets now: https://pypi.org/project/jupyterlab-widgets/.

mbektas: > There is a federated extension for ipywidgets now: https://pypi.org/project/jupyterlab-widgets/.

thanks for bringing this up. I am not sure if we can integrate ipywidgets to desktop app using that python package. For the desktop app, all js front-end components are bundled with the electron-app and the conda environment acts as the backend/server of the desktop app. that's why everything on the front-end needs to be pre-bundled using webpack (https://github.com/jupyterlab/jupyterlab_app/blob/master/webpack.browser.js and https://github.com/jupyterlab/jupyterlab_app/blob/master/src/browser/extensions/index.ts). Please let me know if there is a better way to bundle the ipywidgets js package.

jasongrout: I think compiling in ipywidgets to start with is great (and a sign that leaving in the ability to compile things into jlab was important).

It would be interesting to explore how you might be able to take advantage of prebuilt extensions. Presumably your core webpack bundle can load prebuilt extension bundles just as well as our core webpack bundle.

blink1073 (approved):
Thank you!

Release.md
mbektas: Thank you @jasongrout . I will be researching that. We will need it to support jlab extensions in the App as well.

PR #245, opened 2021-09-24
Add missing extensions

### LINKED ISSUE #244: Could we have Find?
### Problem

I cannot find the Find in the menu.  The Browser version has it, perhaps because the Browser has it.

krassowski:
It seems that some new extensions were added to imports and dependencies but not to the extensions list in the browser. Should fix #244 (I hope).


krassowski:
There is also `@jupyterlab/docprovider-extension` but I did not add it as it is for RTC only I think...

mbektas:
thank you @krassowski !

mbektas (approved):
LGTM!

PR #250, opened 2021-09-26
Update README.md

manojanasuri16:
Added Arch Linux Installer


krassowski (changes requested):
Thanks for starting the work on this! I believe it would address #251, but alternatively I understand we could just distribute a `snap` or `flatpak` as a cross-platform installer (if this does not sound preferable to you please engage in the discussion in #251).

As for this pull request, I believe we would want to have the CI job generating the installer in this   repository and avoid linking to external resources to:
1) ensure security and integrity of the installers in the long term (while I do believe in your good intent, it is just easier to monitor what is going on in CI jobs on organization accounts than on private accounts).
2) ensure that it stays up-to-date when we update the app

mbektas:
We can only provide links to CI output of this repository. Please raise a PR with code changes only.

Issue #242, opened 2021-09-24
JavaScript output is disabled in JupyterLab app


djinnome:
Hi folks

## Description
![Screen Shot 2021-09-24 at 1 40 09 PM](https://user-images.githubusercontent.com/144210/134737169-1955e8fd-22f2-46c5-a62c-eefefa2f8e89.png)

## Reproduce

1. Start JupyterLab app
2. In any code cell, type `%matplotlib inline` or any other command that uses javascript.
3.  See warning: `JavaScript output is disabled in JupyterLab`

## Expected behavior

Javascript works in notebook and regular Jupyter lab.

## Context

- Operating System and version: Mac OS X Mojave
- Browser and version: Chrome Version 93.0.4577.82 (Official Build) (x86_64)
- JupyterLab version: Version 3.1.12-1

djinnome:
I am still getting the same error message in JupyterLab version: Version 3.1.12-2

Do I need to install JupyterLab app from source to get this change, or am I just doing something wrong?

mbektas:
@djinnome we haven't published a new version with the fix yet, please wait for it (~1-2 days). otherwise you can build from master by following the instructions in [README.md](https://github.com/jupyterlab/jupyterlab_app/blob/master/README.md).

djinnome:
Confirmed that the fix works in Version 3.1.13-1 (3.1.13-1) 

However, when I tried to copy the version number in the "About jupyterlab_app" window, I got a pop-up window with the following error message:

**A JavaScript error occurred in the main process**

```
Uncaught Exception:
TypeError: Cannot read property 'copy' of null at KeyboardShortcutManager.copy 
(/Applications/JupyterLab.app/Contents/Resources/app.asar/build/out/main/shortcuts.js:57:55)
```

PR #254, opened 2021-09-26
Automating JupyterLab version updates

mbektas:
- added script to update JupyterLab version
- updated instructions

fixes https://github.com/jupyterlab/jupyterlab_app/issues/220
fixes https://github.com/jupyterlab/jupyterlab_app/issues/242

### DIFF
--- a/Release.md
+++ b/Release.md
@@ -30,8 +30,6 @@ In order to change the JupyterLab version bundled with the App:
 
 3. Update `jupyterlab` and `ipywidgets` python package versions in [`env_installer/construct.yaml`](env_installer/construct.yaml)
 
-4. If new extension packages were added (or removed) update the constant with the list of `extensions` in [`src/browser/extensions/index.ts`](src/browser/extensions/index.ts)
-
 Note that after updating the bundled JupyterLab version, it is necessary to bump JupyterLab App version using `tbump` as described in the section below. Run `check_version_match` script before committing the changes to ensure version integrity.
 
 ```bash
--- a/Release.md
+++ b/Release.md
@@ -20,9 +20,21 @@ If the JupyterLab version is not changing with the new JupyterLab App release th
 
 In order to change the JupyterLab version bundled with the App:
 
-1. Update all the dependencies in [package.json](package.json) (packages with `@jupyterlab` scope and also `@jupyter-widgets/jupyterlab-manager` for ipywidgets). Versions of `@jupyterlab` packages can be found in the commits for the release tag at https://github.com/jupyterlab/jupyterlab/releases
+1. Update all `@jupyterlab` package dependencies in [package.json](package.json) using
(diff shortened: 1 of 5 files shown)


Release.md
```diff
@@ -20,9 +20,21 @@ If the JupyterLab version is not changing with the new JupyterLab App release th
...
-1. Update all the dependencies in [package.json](package.json) (packages with `@jupyterlab` scope and also `@jupyter-widgets/jupyterlab-manager` for ipywidgets). Versions of `@jupyterlab` packages can be found in the commits for the release tag at https://github.com/jupyterlab/jupyterlab/releases
+1. Update all `@jupyterlab` package dependencies in [package.json](package.json) using
+    ```bash
+    yarn set_jupyterlab_version <new-jlab-version>
+    ```
+    `<new-jlab-version>` must match a released JupyterLab version such as `3.1.13`. This command will update dependencies with `@jupyterlab` scope.
+
+2. Update `@jupyter-widgets/jupyterlab-manager` version in [package.json](package.json) for ipywidgets if a compatible newer version is available.
+
+3. Update `jupyterlab` and `ipywidgets` python package versions in [env_installer/construct.yaml](env_installer/construct.yaml)
```
krassowski, line 31:
```suggestion
3. Update `jupyterlab` and `ipywidgets` python package versions in [env_installer/construct.yaml](env_installer/construct.yaml)

4. If new extension packages were added (or removed) add them to the `const extensions` in `src/browser/extensions/index.ts`
```

blink1073 (commented):
Maybe add the handling of the index file to the script?

mbektas:
> Maybe add the handling of the index file to the script?

Automating that part is quite hard since `package.json` usually has more `@jupyterlab` items than `extensions` list. We can detect if an extension in extensions list is not in package.json, but `yarn build` will usually fail in that case already. Please let me know if I am missing something.

blink1073:
I would say include everything in [`jupyterlab.extensions`](https://github.com/jupyterlab/jupyterlab/blob/776fee997d4ed8596989872e6a11e7354cc7732e/jupyterlab/staging/package.json#L209) and maybe have an explicit exclude list for ones we don't want, to avoid situations like https://github.com/jupyterlab/jupyterlab_app/pull/245

krassowski:
I guess the same goes for [`jupyterlab.mimeExtensions`](https://github.com/jupyterlab/jupyterlab/blob/776fee997d4ed8596989872e6a11e7354cc7732e/jupyterlab/staging/package.json#L248). I just noticed that we are missing `@jupyterlab/javascript-extension` (which would explain #242) and `@jupyterlab/vega5-extension`.

blink1073 (commented):
This instruction needs updating

blink1073 (approved):
Thank you!

PR #255, opened 2021-09-26
Add fullscreen mode (F11)

### LINKED ISSUE #246: Can we use it in full screen mode?
I am a fan of Jupyter Lab when I work in browser full screen mode (pressing F11).

I have just trying it out (Desktop app) - I couldn't find full screen option there, I was only able to manage to hide the taskbar but the title bar was still there. Any way to make it full screen?

krassowski:
Fixes #246. When leaving the full screen mode the menu bar was appearing for me which is a bug in Electron (https://github.com/electron/electron/issues/20237), which according to comments only affect Linux and was supposedly fixed in 7.0 but I still see it in 10.x. I added an innocent workaround in 12819f95f7ebd8f0f155fdb618896f1cc353ef01.

### DIFF
--- a/src/main/shortcuts.ts
+++ b/src/main/shortcuts.ts
@@ -27,6 +27,11 @@ namespace IShortcutManager {
     let zoomEvent: AsyncRemote.IEvent<void> = {
         id: 'KeyboardShortcutManager-zoom'
     };
+
+    export
+    let fullscreenToggledEvent: AsyncRemote.IEvent<void> = {
+        id: 'KeyboardShortcutManager-fullscreen'
+    };
 }
 
 /**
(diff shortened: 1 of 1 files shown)


krassowski:
I only tested it on Linux. There is some custom logic listening to leaving full-screen for Mac (which I do not understand, sorry not a Mac user) so it might benefit from testing by a Mac user (the installers should be available on CI soon).

mbektas:
Thanks @krassowski . I tested on Mac but F11 key didn't have any effect for app since it is a system shortcut. Electron's default full screen shortcut is `Ctrl + Cmd + F` on Mac and that has been working already.

mbektas (approved):
LGTM!

hellojinwoo:
On windows 10, `F11` works!

PR #258, opened 2021-09-28
Upgrade to JupyterLab 3.1.13

mbektas:
- using absolute jupyterlab package versions to enable old version builds
- convert `fs-extra` usage to `fs` since it is dev dependency
- update documentation with`%pip install`
- upgrade to jupyterlab 3.1.13


blink1073 (approved):
Thanks!

PR #273, opened 2021-10-03
Launch with command-line or click

mbektas:


blink1073 (approved):
Nice, thank you!

krassowski:
Should we discuss the name of the command, this is whether this should be `jlab`, `jupyterlab`, or something else like more explicit `jupyterlab-desktop`? I saw users complaining about being confused with the variations on Windows (`jupyter lab`, `jupyter-lab`) already. @jasongrout any thoughts?

blink1073:
👍🏼 for naming the executable `jupyterlab-desktop`, which the user could alias to something shorter if they want.

mbektas:
command name is open for discussion. Here is what my thinking with jlab was: I wanted to have the shortest meaningful name with `jlab`. It is a name we often use to refer to JupyterLab in conversations. A name like `jupyterlab-desktop` is more explicit but user needs to type till `jupyterl` for tab completion to work if they have `jupyter` command already on the same environment.

jasongrout:
My two cents: I kind of like `jlab`:

* Mehmet's right that it is short and sweet. It's nice that it is very functional out of the box.
* I think it probably shouldn't be `jupyter-something`, so that it won't conflict with our namespace of for binaries invoked as a jupyter subcommand (like `jupyter something`). To me, it seems that `jupyterlab-desktop` is to confusing with `jupyter-desktop` (if that ever happens).
* My guess is that those downloading and using this are looking for a prebuilt solution-in-a-box, and aren't looking to immediately alias it to have something more functional.

But I'm also the person advocating for `micromamba` to be renamed to something like `mm` on the command line because it's much easier to type (in the end, I made an alias in my `.zshrc`, so maybe that's a vote in favor of Steve's suggestion, but at the same time, there are probably lots of people wanting to use this easily from the command line that would rather not muck around in config files to create an alias, or figure out how to do an alias on Windows).

Besides, what else do you think we should reserve `jlab` for? It's a prime name, but this seems like a prime opportunity as well.

PR #294, opened 2021-10-10
release v3.1.18-1

mbektas:


krassowski:
There is indeed #296, which can be solved by changing the default installation path, or worked around by adding "if you had jupyterlab-app installed previously, please uninstall it before upgrading to jupyterlab-desktop" to the release notes.

mbektas:
> There is indeed #296, which can be solved by changing the default installation path, or worked around by adding "if you had jupyterlab-app installed previously, please uninstall it before upgrading to jupyterlab-desktop" to the release notes.

I see. I think we should prepare uninstall instructions and place a link to it in release notes. Windows uninstallation is two step process. let me work on that.

PR #311, opened 2021-10-23
Set up automated JupyterLab release syncing

krassowski:
This will check for new JupyterLab releases every day at 17:30 UTC and if it finds a new version it will open a PR with appropriate updates. It will look like this one: https://github.com/krassowski/jupyterlab_app/pull/5

### DIFF
--- a/.github/workflows/sync_lab_release.yml
+++ b/.github/workflows/sync_lab_release.yml
@@ -1,0 +1,71 @@
+name: Check for new JupyterLab releases
+
+on:
+  schedule:
+    - cron: 30 17 * * *
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  check_for_lab_updates:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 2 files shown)


.github/workflows/sync_lab_release.yml
```diff
@@ -0,0 +1,71 @@
...
+            yarn install
+          fi
+      - name: Create a PR if needed
+        shell: bash
+        env:
+          GITHUB_USER: ${{ secrets.GITHUB_USER }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -eux
+          # if resulted in any change:
```
jtpio, line 56: Wondering how this would work if we wanted to start using pre-release packages?

For example we might want to start using the upstream `4.0.0-rc.x` packages before the final release, and ship pre-releases of the desktop app. 

Looks like this script would try to downgrade to the stable release?

krassowski: We could condition the use of release/pre-release versions on branch fairly easily. That we are playing a catch up game here and still have a lot of legacy code so I don't know how soon we can start testing pre-releases.

PR #317, opened 2021-10-29
Custom python environment support

mbektas:
Adding basic support for custom Python environment installed on user machine
- Python environment info is shown on status bar
- conda and venv are both supported
  -  Status bar item has the label as follows: `conda: env_name` (for conda), `venv: env_name` (for virtualenv), `system: python` (for other cases)
  - Hovering on status bar item shows env type and name, python path, python & jupyterlab versions
- Clicking status bar item launches the environment select dialog
- Setting is persisted as part of application state
- At launch time, compatibility of selected python path is checked (including bundled environment, since user can modify it) and if found incompatible, Python path selector dialog is shown with the incompatibility error message.

Status bar item
env-status-bar

Changing environment
change-env

Failed launch due to incompatible environment
python-incompatible


src/main/app.ts
```diff
@@ -326,6 +382,100 @@ class JupyterApplication implements IApplication, IStatefulService {
...
+                        <label for="bundled">Use bundled Python environment</label>
+                    </div>
+                    <div class="row">
+                        <input type="radio" id="custom" name="env_type" value="custom" ${!useBundledPythonPath ? 'checked' : ''} onchange="handleEnvTypeChange(this);">
+                        <label for="custom">Use custom Python environment</label>
+                    </div>
+
+                    <div class="row">
+                        <div style="flex-grow: 1;">
+                            <input type="text" id="python-path" value="${pythonPath}" readonly style="width: 100%;"></input>
```
krassowski, line 427: https://xkcd.com/327/

package.json
```diff
@@ -258,6 +259,7 @@
     "@types/semver": "^7.3.4",
     "@types/yargs": "^16.0.0",
     "bottlejs": "^2.0.0",
+    "ejs": "^3.1.6",
```
krassowski, line 262: I was thinking more `renderToString` style from `react-dom` (which already is a dependency), but I guess that would work too. I don't think `yarn.lock` was updated, how much does inclusion of `ejs` costs us in terms of extra dependencies?

mbektas: ejs was already included through another dependency. I will also check out react-dom renderToString you mentioned.

bollwyvl:
Hm, it looks like my reply was swallowed by the aether (happens), here's a paraphrase:

Strictly from a `conda` perspective: if the canonical `conda activate` (or shell-appropriate analog) is not applied, the eventual environment will be inconsistent, creating weird problems.... known offenders include java/clojure, or anything that has a `pre-activate` script.

Some approaches:
- dynamically make a shell-appropriate (`.sh` or `.bat`) wrapper script that `activate`s and then launches 
  - gross, slow...
    - ...but works every time
- use `conda run`
  - clean, relatively fast
    - but doesn't work so well with `mamba`... 
      - ...unless the "heavyweight" shell initialization is in place, which has its own problems
      - ...much less `venv`

At any rate, perhaps an appropriate play is to handle the two incumbent providers explicitly, implemented as the former, and then a _Customize Template..._ which would allow someone to further tweak them, or replace entirely with e.g. `nix`, `vagrant`, some docker abomination, etc. (many of which `conda` would be sufficient to provision, a la as discussed https://github.com/jupyterlab-contrib/jupyterlab-contrib.github.io/issues/23#issuecomment-956508533

package.json
mbektas: @krassowski I looked into ReactDOMServer.renderToString. it doesn't look like an easy integration for our use case. we generate the HTML of the dialogs on the app backend using node.js. Dialog front-ends are quite basic HTML and JS. We would need to use react for dialog front-ends which would be significant overhead just for sanitize purposes. Also, sanitization would happen on the front-end which may not be as secure as intended.

Please let me know if I missed something and if you see an easier integration of ReactDOMServer.renderToString.

mbektas:
> At any rate, perhaps an appropriate play is to handle the two incumbent providers explicitly, implemented as the former, and then a _Customize Template..._ which would allow someone to further tweak them, or replace entirely with e.g. `nix`, `vagrant`, some docker abomination, etc. (many of which `conda` would be sufficient to provision, a la as discussed [jupyterlab-contrib/jupyterlab-contrib.github.io#23 (comment)](https://github.com/jupyterlab-contrib/jupyterlab-contrib.github.io/issues/23#issuecomment-956508533)

@bollwyvl thanks for bringing this up. conda run looks pretty nice but venv requires a separate solution too. We should go for creating launch scripts that take care of env activation and then launching jlab as you suggested. As far as I could see, our packages in the bundled environment don't have pre-activate/activate scripts and this case seems to be rare. So, I think we can implement launch script generation in another PR.

src/main/app.ts
```diff
@@ -323,6 +403,124 @@ class JupyterApplication implements IApplication, IStatefulService {
...
+                const selectPythonPathButton = document.getElementById('select-python-path');
+
+                function handleSelectPythonPath(el) {
+                    ipcRenderer.send('select-python-path');
+                }
+
+                function handleReleasesLink(el) {
+                    ipcRenderer.send('launch-installer-download-page');
+                }
+
```
krassowski, line 490: Copied over from the other one I guess?

```suggestion

```

src/main/app.ts
```diff
@@ -266,6 +305,36 @@ class JupyterApplication implements IApplication, IStatefulService {
             shell.openExternal('https://github.com/jupyterlab/jupyterlab-desktop/releases');
         });
 
+        ipcMain.on('select-python-path', (event) => {
+            dialog.showOpenDialog({
+                properties: ['openFile', 'showHiddenFiles', 'noResolveAliases'],
```
krassowski, line 310: - Can we add `defaultPath` pointing to the last recently used path (if any)? Changing between pyenv environments would be easier this was.
- Should we add `filters` to only show python executables? [Not sure of this one]

src/main/app.ts
```diff
@@ -323,6 +403,124 @@ class JupyterApplication implements IApplication, IStatefulService {
...
+        if (pythonPath === '' || pythonPath === bundledPythonPath) {
+            useBundledPythonPath = true;
+        }
+        const configuredPath = pythonPath === '' ? bundledPythonPath : pythonPath;
+        const requirements = this._registry.getRequirements();
+        const reqVersions = requirements.map((req) => `${req.name} ${req.versionRange.format()}`);
+        const reqList = reqVersions.join(', ');
+
+        const message = reason === 'change' ?
+            `Select the Python executable in the conda or virtualenv environment you would like to use for JupyterLab Desktop. Python packages in the environment selected need to meet the following requirements: ${reqList}.` :
```
krassowski, line 438: Either here or in documentation we should mention that this changes which prebuilt extensions get loaded.

```suggestion
            `Select the Python executable in the conda or virtualenv environment you would like to use for JupyterLab Desktop. Python packages in the environment selected need to meet the following requirements: ${reqList}. Please note that prebuilt extensions installed in selected environment will be loaded.` :
```

src/main/app.ts
mbektas: yep, good catch!

src/main/app.ts
mbektas: I see, agreed on `defaultPath`. not sure about filters, since users might have aliases etc.

krassowski (approved):
LGTM, thank you!

PR #318, opened 2021-10-30
macOS code signing

mbektas:


mbektas:
moved to https://github.com/jupyterlab/jupyterlab-desktop/pull/428

Issue #269, opened 2021-09-30
There is no "Enable Extensions Manager" !


NILICK:
I installed latest version of JupyterLab App v3.1.13-1 but there is not "Enable Extensions Manager" option in setting menu.
I then run `jupyter labextension install @jupyter-widgets/jupyterlab-manager` but I have same problem. 

```
JupyterLab v3.1.13
/opt/JupyterLab/resources/jlab_server/share/jupyter/labextensions
        @bokeh/jupyter_bokeh v3.0.2 enabled OK (python, jupyter_bokeh)
        @pyviz/jupyterlab_pyviz v2.1.0 enabled OK (python, pyviz_comms)
        @jupyter-widgets/jupyterlab-manager v3.0.1 enabled OK (python, jupyterlab_widgets)

Other labextensions (built into JupyterLab)
   app dir: /opt/JupyterLab/resources/jlab_server/share/jupyter/lab
        @ijmbarr/jupyterlab_spellchecker v0.7.1 enabled OK
        @jupyter-widgets/jupyterlab-sidecar v0.6.1 enabled OK
        jupyter-leaflet v0.14.0 enabled OK
```
![Screenshot from 2021-09-30 18-55-27](https://user-images.githubusercontent.com/65063089/135484786-42dc5cbb-3472-4e71-868c-62c8c05199ff.png)

mbektas:
Installing additional JupyterLab Extensions are not supported in the desktop application at the moment. We are considering adding this support in a future release.

krassowski:
Please note that prebuilt extensions are now supported since 3.2.1-2 and can be installed via `pip` or `conda`, but extension manager is still disabled as it does not support installation of prebuilt extensions yet.

hellojinwoo:
@krassowski If the installation is possible with the command `jupyter labextension install`, can't I install the extension? For example, the jupyterlab theme `city-lights-theme`, the installation command is as follows:

`jupyter labextension install @yudai-nkt/jupyterlab_city-lights-theme
`

Can't I install this theme on the Jupyterlab Desktop app because it is not installed via `pip` nor `conda`? 

Theme repo: https://github.com/yudai-nkt/jupyterlab_city-lights-theme

krassowski:
[...]

PR #322, opened 2021-10-31
Port to federated extensions (with eager loading of shared extensions)

### LINKED ISSUE #49: Lab extensions
We should figure out an elegant way to handle lab extensions in the application. 

If were to use "jupyter labextension install <extensions>", the JupyterLab application assets are built and run in the application directory. So, if we wanted to use this command as-is with the Electron application, we really would only be using the extension tarballs created by "npm pack" and the settings. Then, we'd still have to use npm to install the extensions for the Electron app and we're basically repeating the process, or something similar to it. This would also mean that any lab extension installed in the standard JupyterLab would be installed for the Electron app. And this also requires the user to have npm installed to install a lab extension. 

Any thoughts on the best way to approach this?

### LINKED ISSUE #264: Cannot use git extension on jupyterlab_app, even though I can see the tab.
image

Even though I have installed jupyter_git and used it on usual jupyter lab environment, I cannot use it on jupyterlab_app. 
I can see the tab for `git`, but when I click the tab, it shows only help. 

Plus, why can't I see the extension on the left side bar in the jupyterlab_app?

### LINKED ISSUE #288: Plotly figures don't show up
Hello,
using the latest version of this application on Ubuntu 20.04, the plotly figures seems not to work.
Here is an example:
( the output of the figure is empty).

image

### LINKED ISSUE #323: Do we need prefix free?
Our template currently includes:

```
<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/prefixfree/1.0.7/prefixfree.min.js"></script>
```

But this was added years ago. It is very likely that we can drop it.

krassowski:
Fixes #288, closes #49 and fixes #264. Does not solve #269, i.e. the extension manager remains disabled as we do not offer support for source extensions but currently the extension manager assumes that both prebuild (federated) and source extension are supported.

Reverts #316 - implements a proper fix by using `<base href="${juypter_server_url}">`, which also allows to use shipped MathJax over the CDN version.

Closes #323.

### Code changes
- uses webpack module federation, in line with upstream and retrolab to deliver the core packages to the frontend
   - differently than upstream, the core packages are loaded eagerly which reduces the extent of changes needed; this might be suboptimal for web deployments but it should not matter for a desktop app that much
- properly re-uses the `PageConfig` provided by the server; this also means that we get the URL for the licenses endpoint
- the lab app will now be loaded asynchronously after the prebuilt extensions are fetched; I kept the changes in the App to the minimum (i.e. I kept the old synchronous methods) to avoid conflicts with other work you may have and used `then` instead.
- updated target to `es2020` since we now ship Electron 12 (2019 would probably do fine too)
- we now copy the resolutions from upstream as those define the version of shared packages we offer; we could create our own resolutions, but I believe it is better to stick to the upstream algorithm and ensure we are in sync

Please note that our custom extensions (`electron-extension`, `desktop-extension`) are not shared, i.e. third-party extensions cannot depend nor use them.

### User-facing changes

1) Local MathJax copy will be used, which means faster load time and ability to render math when offline
2) The licences component is now available (see the last screenshot) as we have reuse upstream `PageConfig` which has the required endpoint URL included
3) Prebuilt extensions, such as MIME renderers and other output-related extensions work if installed into the correct environment

plotly:

Screenshot from 2021-10-31 12-48-29

jupyterlab-git:

Screenshot from 2021-10-31 13-24-25

jupyterlab-lsp + pylsp server:

Screenshot from 2021-10-31 13-37-00

licences component

Screenshot from 2021-10-31 12-54-56


meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/how-to-run-jupyter-lab-without-the-terminal-on-windows-10/11419/12

krassowski:
New user-facing change with the latest commit, this is the `appName` in `PageConfig` is now set to `JupyterLab Desktop`: 

desktop

krassowski:
I am not fully convinced about the name change; the longer name might help in user awareness and support, but is well, longer. Either would work fine for me - please let me know what you think.

mbektas (approved):
LGTM. Thanks!

scripts/buildutil.js
```diff
@@ -251,6 +251,9 @@ if (cli.flags.setJupyterlabVersion !== "") {
                     }
                 }
 
+                // copy resolutions as is
+                pkgjsonFileData.resolutions = newResolutions;
```
krassowski, line 255: Needs the same for `singletonPackages`.

blink1073 (approved):
Amazing, thank you!

jtpio:
Really nice!

PR #329, opened 2021-11-04
Add `ipympl` to default environment

### LINKED ISSUE #354: Bundle ipympl
## Description

The `%matplotlib widget` magic fails out of the box

## Reproduce

1. Install JupyerLab Desktop
2. Set `%matplotlib widget`
3. Try to plot something
4. See error `ModuleNotFoundError: No module named 'ipympl'`

## Expected behavior

Plot

## Workaround

`%pip install ipympl`
and restart Jupyter Lab Desktop

## Context

- Operating System and version: Windows 10
- JupyterLab-Desktop version: 3.2.4-3
- Installer: .exe

ianhi:
This allows for interactive matplotlib plots to work out of the box which is key for feature parity with classic notebook (where matplotlib itself provides the widget)

This came up in https://github.com/matplotlib/ipympl/issues/388 where someone expected `ipympl` to be installed but found that it was not.

Closes #354

### DIFF
--- a/env_installer/construct.yaml
+++ b/env_installer/construct.yaml
@@ -22,6 +22,7 @@ specs:
   - conda
   - ipywidgets 7.6.5
   - jupyterlab 3.2.4
+  - ipympl >= 0.8.2
   - matplotlib-base
   - numpy
   - pandas


env_installer/construct.yaml
```diff
@@ -21,6 +21,7 @@ license_file: ../LICENSE
 specs:
   - conda
   - ipywidgets 7.6.5
+  - ipympl 0.8.2
```
krassowski, line 24: +1 seems reasonable to me.
Do we want/need the version pin?

ianhi: Nope! It should be at least `>=0.8.2` but the spaces confused me because it's different than the syntax a normal conda env file?

krassowski: https://github.com/conda/constructor/blob/master/CONSTRUCT.md#specs says that normal conda env sntax should work. So I guess `ipympl >=0.8.2` should do it.

mueslo: does conda support version statements like `ipympl ~=0.8.2` ?

https://www.python.org/dev/peps/pep-0440/#compatible-release

ianhi: I can say pretty confidently that for ipympl you probably want `>=`. The version numbers there pretty much only reflect improvements to the widget and don't reflect compatibility with jupyterlab

ianhi: but good thought! and probably relevant in the future

krassowski (approved):
Thanks!

mbektas (approved):
Thanks!

Issue #238, opened 2021-09-23
Jupyter Server Not Found (executable not in PATH)


alejandroschuler:
## Description

When starting jupyterlab_app, I get the message:

> Jupyter Server Not Found
> We were unable to launch a Jupyter server, which is a prerequisite for JupyterLab Native. If Jupyter is installed as a python module, but the python executable is not in your PATH, specify the executable location below. Otherwise, try installing or updating Jupyter. The Jupyter notebook version must be 4.3.0 or greater.

Despite a) having a working jupyter-lab installation that I can spin up from the terminal and which appears in the browser, b) having the python executable in my path, and c) having the requisite version of jupyter. 

When I manually point jupyterlab_app to the python executable (`/Users/aschuler/.pyenv/shims/python`) via the `CHOOSE PATH` it spins the icon again for a moment and then returns the same "Server Not Found" message.

## Context

I am running python through **pyenv**, which may be the issue here. Relevant information:
```
aschuler@Alejandros-MacBook-Air ~ % pyenv versions
  system
* 3.9.6 (set by /Users/aschuler/.pyenv/version)
```
```
aschuler@Alejandros-MacBook-Air ~ % echo $PATH    
/Users/aschuler/.pyenv/shims:/Applications/Sublime Text.app/Contents/SharedSupport/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin
```
```
aschuler@Alejandros-MacBook-Air ~ % jupyter --version
jupyter core     : 4.7.1
jupyter-notebook : 6.4.0
qtconsole        : 5.1.1
ipython          : 7.26.0
ipykernel        : 6.0.3
jupyter client   : 6.1.12
jupyter lab      : 3.1.13
nbconvert        : 6.1.0
ipywidgets       : 7.6.3
nbformat         : 5.1.3
traitlets        : 5.0.5
```
```
aschuler@Alejandros-MacBook-Air ~ % which python
/Users/aschuler/.pyenv/shims/python
```
```
aschuler@Alejandros-MacBook-Air ~ % which jupyter-lab
/Users/aschuler/.pyenv/shims/jupyter-lab
```

(MK): Extra details from title: jupyterlab_app fails to start on MacOS 11.5.2 with  (pyenv issue?)

alejandroschuler:
[...]

PR #339, opened 2021-11-18
Improve error reporting and use pre-defined token

krassowski:
https://github.com/jupyterlab/jupyterlab-desktop/issues/238 demonstrates that for an unknown reasons the Jupyter Server does not get setup correctly for some users. It is very difficult to debug currently, so I tried to address this in multiple ways:
- removing one of the modes of failures which was in retrieving the token (which if not found could lead to promise rejection and error screen state)
- displaying the actual error that led to the error screen (and encouraging reporting of these errors)
- allowing users to change environment with the selector introduced in https://github.com/jupyterlab/jupyterlab-desktop/pull/317
- rewriting the error message which was incorrectly suggesting that it was due to server not being in path, where in reality there are several code paths which could lead to the promise rejection and that screen; instead it is now more accurately titled "Jupyter Server Initialization Failed"
- I also removed the "Install Jupyter" link, because the JupyterLab Desktop now ships with a builtin conda environment so it is redundant and it was directing users to rather not very useful generic Jupyter website (so they could end up installing Notebook and getting frustrated that it did not help).
- I added code for parsing the actual version of Jupyter Server that got started, but end up not using it for now. We should consider if we want to add a check on the version at this level too - it could help against some inconsistent environments - but it can wait for another PR.
- The server standard error output is now logged into the console. We may want to change this in the future (or make it optional, only for debugging) but it seems important for this stage of development.

I also restored the XSRF checks as I do not see any reason for having those disabled when the token is properly set up.

Tested by manually adding a `reject(new Error('test'))`:

Screenshot from 2021-11-18 21-00-50

And selecting a non-Python path fusing the first button:

Screenshot from 2021-11-18 21-00-56

The environment selector works well too:

Screenshot from 2021-11-18 21-01-22

### DIFF
--- a/src/browser/components/error/servererror.tsx
+++ b/src/browser/components/error/servererror.tsx
@@ -10,6 +10,8 @@ namespace ServerError {
     export
     interface Props {
         launchFromPath: () => void;
+        changeEnvironment: () => void;
+        error: Error
     }
 }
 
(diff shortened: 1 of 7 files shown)


krassowski:
The CI errors are not related to this PR, but to lack of peer dependency pinning. Yarn is installing the latest versions of peer dependencies which causes compilation issues. Merging https://github.com/jupyterlab/jupyterlab-desktop/pull/337 will temporarily solve this issue, but we should find a way to better locking of peer dependencies.

src/browser/components/error/servererror.tsx
```diff
@@ -18,14 +20,24 @@ function ServerError(props: ServerError.Props) {
...
+                    Jupyter Server, which is a prerequisite for JupyterLab Desktop, did not initialize properly.<br/>
+                    This might be because of an issue with the selected environment. You can change environment or try setting path to the server using buttons below.<br/>
+                    </p>
+                    <p>
+                    If this does not seem right, please report this issue in the JupyterLab Desktop repository, providing the error details presented below:
+                    </p>
+                </span>
+                <pre className='jpe-ServerError-error'>{props.error.name}: {props.error.message}</pre>
                 <div className='jpe-ServerError-btn-container'>
                     <button className='jpe-ServerError-btn' onClick={props.launchFromPath}>CHOOSE PATH</button>
```
mbektas, line 36: `launchFromPath` is not actually applying the selected path anymore. we should remove `CHOOSE PATH` option from here which was supposed to do what `CHANGE ENVIRONMENT` does anyway.

krassowski: Removed in 7ebd7fa6658bc8352b317d3e1ce885f2adb7ae32. Also rebased so it should build smoothly now.

mbektas (approved):
Thank you!

PR #341, opened 2021-11-19
Upgrade Electron to 13.6.2

krassowski:
New version of electron (16.0) was just released and 12.x is no longer supported. This PR updates to 13.x which is still supported to give us some more time to upgrade to modern versions.


mbektas (approved):
Thanks!

Issue #206, opened 2021-09-15
Copy, cut and paste with context menu


krassowski:
### Problem

First, thanks for the awesome work on reviving this! I tested on Ubuntu and it works!

There is the `Shift + Right Click for Browser Menu` entry, but `Shift + Right Click` does not do anything for me. This means that some actions (copy, paste, cut, copy link address) are not accessible now.

![Screenshot from 2021-09-15 17-31-31](https://user-images.githubusercontent.com/5832902/133476306-b9bfaf1c-830d-4c98-977a-8cc13ee314af.png)

### Proposed Solution

- Work with upstream JupterLab to add full self-contained functionality of copy, paste, and cut (which are already available in Text Editor but not in Notebook) and maybe additional actions (copy link) too
- Disable the `Shift + Right Click for Browser Menu` entry in JupyterLab App as it is misleading (this should be possible with the menu schema files).

krassowski:
I explored four approaches:
- A) using `webContents` API (e.g. `webContents.copy`) for both interacting with clipboard and with the page, and communicating via remotes
- B) using `Electron.clipboard` directly for interacting with clipboard, Web APIs like [`Selection`](https://developer.mozilla.org/en-US/docs/Web/API/Selection) and DOM API for interacting with the page, and communicating via remotes
- C) using [`Clipboard API`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) on client side only, no communication with server
- D) adding a fallback context menu with Electron-native implementation of commonly needed actions

The code for A and B would rougly look as follows:

<details>
<summary>Code</summary>

In `desktop-extension`

```ts
app.commands.addCommand(CommandIDs.copySelection, {
    label: 'Copy Text',
    execute: () => {
        const selection = window.getSelection();
        asyncRemoteRenderer.runRemoteMethod(IAppRemoteInterface.writeToClipboard, selection.toString()).catch(console.warn);
    },
    isVisible: () => {
        const selection = window.getSelection();
        return selection.toString() !== '';
    },
});
```

and server side (in `main/app`):

```ts
asyncRemoteMain.registerRemoteMethod(IAppRemoteInterface.writeToClipboard,
    (data: string): Promise<void> => {
        // approach A:
        // clipboard.writeText(data);
        // approach B:
        this._window.webContents.copy();
        return Promise.resolve();
    });
```

</details>

Problems with approaches A and B:
[...]

PR #342, opened 2021-11-20
Add (temporary?) fallback context menu

### LINKED ISSUE #297: Shift + Right doesn't work
When I right click, I see in the bottom of the menu that I need to `Shift ` + `Right `click to bring up the browser menu.
image

However it doesn't seem to work with this application. When I `Shift` + `Right` Click nothing happens.

(I found that this feature is implemented and no issue was reported here: https://github.com/jupyterlab/jupyterlab/issues/3224 )

krassowski:
Fixes #297 and implements the "Approach D" described in https://github.com/jupyterlab/jupyterlab-desktop/issues/206#issuecomment-974670327 (until a better solution, "approach C" is implemented in core JupyterLab).

Please see https://github.com/jupyterlab/jupyterlab-desktop/issues/206#issuecomment-974670327 for technical discussion.

Notebook for testing: https://gist.github.com/krassowski/9cecfcc0efd1177c124a066f7d0ece8b

### User-facing changes

The fallback context menu is available when holding `Shift`:

demo

### DIFF
--- a/src/main/sessions.ts
+++ b/src/main/sessions.ts
@@ -2,7 +2,7 @@
 // Distributed under the terms of the Modified BSD License.
 
 import {
-    BrowserWindow, ipcMain, app
+    BrowserWindow, ipcMain, app, Menu, MenuItemConstructorOptions, clipboard
 } from 'electron';
 
 import {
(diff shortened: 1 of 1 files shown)


mbektas (approved):
great! thanks!

PR #340, opened 2021-11-19
Fix navigation to anchor (location.hash)

krassowski:
Fixes #332 and prevents navigating away to any external website.

PR #344, opened 2021-11-20
Enhance navigation guard to fix hash navigation

krassowski:
Follow-up to #340 - it appears that the more cautious version that I committed did not solve the problem well. This is more strict (forbidding any navigation, which is needed to guarantee that the plain JupyterLab version won't get loaded) and also ensuring that hash navigation works without any prompt.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -3,7 +3,6 @@ import {
 } from 'electron';
 
 const Bottle = require('bottlejs');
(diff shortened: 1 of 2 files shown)


krassowski:
The failure on CI was weird:

> Error: Input required and not supplied: repo_token

Kicked CI

mbektas (approved):
LGTM!

PR #347, opened 2021-11-24
Support traitlets versions < 5.0

krassowski:
Such as >=4.2.1 required by `jupyter_server`. The alternative is to add traitlets>=5.0 as a dependency but I think this solution would is more user-friendly.

This should fix the error from https://github.com/jupyterlab/jupyterlab-desktop/issues/238#issuecomment-977173805. Not sure if it will solve the issue for everyone.


mbektas (approved):
Thanks!

Issue #226, opened 2021-09-22
Displaying available kernels


psychemedia:
Hi

Just trying a first run on a Mac and I noticed a couple of things:

- JupyterLab-app opened with a couple of open notebooks that matched notebooks I had opened using a local notebook running already in browser tabs;
- JupyterLab-app opened into a browser directory on a path that JuypyterLab-app shouldn't have known about;
- multiple kernels are shown as notebook opening options JupyterLab-app launcher:
![image](https://user-images.githubusercontent.com/82988/134421515-242f2d73-e9a6-47d9-917c-258c53c3524e.png)

The *Python 3 (ipykernel)* kernel opens a kernel that seems to be the bundled kernel (`import site; site.getsitepackages()` gives `['/Applications/JupyterLab.app/Contents/Resources/jlab_server/lib/python3.8/site-packages']`. This kernel has the same name as the default kernel in my locally running Jupyter server. In the local browser accessed JupyterLab, the kernel with the same name connects a local kernel (not the JuyterLab-app bundled kernel).

If I have an open notebook in JupyterLab app, when I create a new local server JupyterLab browser window, the notebook open in the app opens in the local browser window (?!) but runs against a different kernel, albeit with the same name:

![image](https://user-images.githubusercontent.com/82988/134422278-97a55588-db58-4f60-93b1-a6fe484dae12.png)

I guess my issue is:

- the shipped kernel name may clash with a local kernel name; which kernel is used depends on which JupyterLab is running (app or local browser accessed server); *this is confusing;*
- the JupyterLab-app and local JupyterLab server seem to be able to share state re: current directory and currently opened notebooks(?!); *this is quite scary and I suspect could lead to all sorts of mishaps depending on what is shared and which server last wrote whatever state is shared);
- I seem to be able to connect to a local R kernel, but trying to connect to other local Py kernels seems to hang the app (that said, my other Py kernels may not be fully working or may be slow to start).

krassowski:
Good points. I think that it largely depends on configuration - if you have kernels installed on system level, any instance of JupyterLab will list those (whether in a virtual environment or not).

> the JupyterLab-app and local JupyterLab server seem to be able to share state re: current directory and currently opened notebooks(?!); *this is quite scary and I suspect could lead to all sorts of mishaps depending on what is shared and which server last wrote whatever state is shared);

[...]

PR #348, opened 2021-11-25
Use a dedicated configuration directory to isolate settings & workspaces

### LINKED ISSUE #240: Launcher Error Failed to construct 'IntersectionObserver': rootMargin must be specified in pixels or percent.
I cannot launch either octave or python kernels.  I get this error:
Launcher Error
Failed to construct 'IntersectionObserver': rootMargin must be specified in pixels or percent.

image

To reproduce it, just install JupyterLab_app on Ubuntu 21.04 and start it.  Then click the Python 3 launcher.

\

## Context
This is on Ubuntu 21.04.  I have an ordinary jupyter-lab that uses the browser installed too, and it works fine.  I used the .deb to install it just now.

\

krassowski:
- Fixes point 1 and point 2 of #226 (making the Desktop app use independent workspaces and independent settings from the system-wide installation)
- Fixes another failure mode causing server startup error (#238) which turned out to be caused by a competing configuration from previous installations: users would have different notebook settings in their home which were masking the settings that we wanted. This is a bit drastic step, so I also added an environment variable to allow to change it back to something else and documented it.
- Edit: this will also close #240 which was a case of a configuration from previous installation breaking JupyterLab Desktop UI
 
I am not sure if the documentation should go to `README` or `User guide` - this will be a prominent change for existing users so maybe README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,20 @@ JupyterLab Desktop sets File Browser's root directory based on the launch method
   - `jlab ../notebooks/test.ipynb` launch notebook with relative path
   - `jlab ../test.py` launch python file with relative path
 
+### Configuration files
+
+By default JupyterLab Desktop will only load and write to Jupyter configuration located in:
+- `%APPDATA%\jupyterlab-desktop` on Windows
+- `$XDG_CONFIG_HOME/jupyterlab-desktop` or `~/.config/jupyterlab-desktop` on Linux
+- `~/Library/Application Support/jupyterlab-desktop` on macOS
+
+ignoring any other Jupyter configuration which may be present in standard Jupyter paths as defined by `jupyter --paths`.
+This includes `jupyter-server` settings, `jupyterlab` settings and workspaces, and any other configuration which would
+normally be shared between Jupyter installation.
+This is necessary to prevent a clash between the Desktop and any previous Jupyter installation.
+
+You can change the configuration path by specifying `JLAB_DESKTOP_CONFIG_DIR` environment variable.
+
 ## Build dependencies
 
 - [conda](https://docs.conda.io)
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

krassowski:
Thanks for the merge! Let's remember to add instructions on copying settings over in the release notes. Draft follows:

> Please note that starting with this version of JupyterLab Desktop the configuration (`jupyter-server` settings, `jupyterlab` settings and workspaces, etc) is no longer shared between previous installations of Jupyter(Lab) and JupyterLab Desktop to avoid workspace leakage, user confusion, and errors such as reported by users who had a pre-existing configuration clashing with a newer version of Jupyter that was shipped with JupyterLab Desktop. You can keep your previous JupyterLab settings by copying them over to the new [configuration path](https://github.com/jupyterlab/jupyterlab-desktop#configuration-files):
> 
> 1. Run `jupyter --paths` to determine where your `config` is stored.
> 2. Find directory called `lab` in one of these paths.
> 3. If there is a `user-settings` directory in `lab`, you can copy the `lab` directory over to the [configuration path](https://github.com/jupyterlab/jupyterlab-desktop#configuration-files) of JupyterLab Desktop to keep your old JupyterLab settings. If there is `workspaces` directory you can decide if you want to keep or remove it.
> 
> For example, on Linux it could (depending on installation) require the following:
> 
> ```
> cp -r ~/.jupyter/lab/ ~/.config/jupyterlab-desktop
> ```
> 
> **Warning**: If you copy over settings from an older major version of JupyterLab (e.g. 2.x) those might cause an error on startup. We only recommend to copy settings over if you already had 3.x installed and are upgrading from previous JupyterLab Desktop version which was using the shared Jupyter configuration.

PR #362, opened 2021-12-03
Allow displaying hidden files

### LINKED ISSUE #361: Cannot display hidden files
## Description

Hidden files require changing server-side configuration, but this is not currently possible.

## Expected behavior

Thanks to https://github.com/jupyterlab/jupyterlab/pull/11206 we can now enable it on server side and allow user to choose in the GUI whether to display the hidden files or not (https://jupyterlab.readthedocs.io/en/stable/user/files.html#displaying-hidden-files).

## Context

- JupyterLab-Desktop version: 3.2.4-3

krassowski:
Fixes #361

Allows to display hidden files:

Screenshot from 2021-12-03 20-55-49

By default hidden files are **not** shown. User needs to go to the `View` menu and check `View Hidden Files`

### DIFF
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -81,7 +81,9 @@ class JupyterServer {
                 '--JupyterApp.config_file_name=""',
                 // use our token rather than any pre-configured password
                 '--ServerApp.password=""',
-                '--ServerApp.allow_origin="*"'
+                '--ServerApp.allow_origin="*"',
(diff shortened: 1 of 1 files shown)


mbektas:
Does this change make `Show Hidden Files` default?

mbektas:
nevermind, I checked the jlab code. that seems to be the case (non-default).

mbektas (approved):
LGTM!

PR #363, opened 2021-12-04
Documentation updates

mbektas:
updating application capture and reordering uninstall instruction sections.


krassowski (commented):
Looks good. Two minor suggestions:

1) we can compress the screenshot saving around 20% (lossless compression)

jupyterlab-desktop

2) we could reduce the margins a little bit in the region where the shadow is not that visible any more, e.g. (this one is compressed too):

suggestion

mbektas:
> Looks good. Two minor suggestions:
> 

sure, I will update it with your image here. I used macOS system tools and they don't seem to support compression.

krassowski (approved):

PR #367, opened 2021-12-11
Release v3.2.5-1

mbektas:


krassowski (approved):
I would add the following to the changelog:
- ship `ipympl` to the default environment, supporting interactive `matplotlib` plots out of the box
- enable displaying hidden files (`View` menu and check `Show Hidden Files`)

PR #368, opened 2021-12-11
Upgrade to Electron 14

krassowski:
This is a proposal to upgrade to Electron 14. It also removes majority of the `TitleBar` code (which I believe is not currently in use).

Electron 14 blog post: https://www.electronjs.org/blog/electron-14-0
We are getting straight to 14.2.2.

Full changelog for 14.0.0: https://www.electronjs.org/releases/stable?page=4#14.0.0

Stack:
- Chromium 91 → 93
- Node v14.17.0 (without changes)
- V8: v9.1 → v9.3

The size of installers increased by less than 1 MB:

| before | after |
|--|---|
| Screenshot from 2021-12-12 00-08-54 | Screenshot from 2021-12-12 00-08-45 |


mbektas (approved):
LGTM. Thanks!

PR #374, opened 2021-12-17
Keep the environment variables when spawning the server

### LINKED ISSUE #373: Environment variables are not loaded
## Description
Hello, when using jupyterlab-desktop the program seems to start in a completely empty environment. Variables such as `$USER` and `$HOME` are unset. This is the case in all notebooks and all terminals.
screenshot_307598
The terminal then sources my `~/.profile` and arrives at nonsense paths
screenshot_583409
(notice how some of them they start with `/.config/` or `/.local/`, since `$HOME` was unset).

## Reproduce
The program was installed from the 3.2.5 [deb](https://github.com/jupyterlab/jupyterlab-desktop/releases/download/v3.2.5-1/JupyterLab-Setup-Debian.deb) via this [aur](https://aur.archlinux.org/packages/jupyterlab-desktop-bin/) package (but with `$pkgver` updated to v3.2.5). It is launched with `/opt/JupyterLab/jupyterlab-desktop`.

Upon first launch I selected the python executable at `/usr/bin/python`, which has the "regular" jupyterlab installed. The packages are working fine (I am able to import everything installed on my system), but there are no environment variables.

## Expected behavior

Environment variables are present. This issue does not exist when I launch the "regular", browser-based jupyterlab in Firefox 95.0.1:
screenshot_048219

## Context

- Operating System and version: Arch Linux
- JupyterLab-Desktop version: 3.2.5
- Installer: .deb

krassowski:
Fixes #373.

Do we want to filter out any environment variables? Looking at https://www.electronjs.org/docs/latest/api/environment-variables there is nothing that we make use of that should not be forwarded to the server. In dev mode I see a lot of variables like:

```
npm_package_dependencies__jupyterlab_rendermime_extension=3.2.5
npm_package_build_directories_buildResources=dist-resources
npm_package_resolutions__jupyterlab_celltags=~3.2.5
npm_package_dependencies__jupyterlab_rendermime_interfaces=3.2.5
npm_package_devDependencies__types_react=~17.0.2
```

which might be just annoying noise (or a great way for debugging).

### DIFF
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -87,6 +87,7 @@ class JupyterServer {
             ], {
                 cwd: home,
                 env: {
+                    ...process.env,
                     PATH: this._registry.getAdditionalPathIncludesForPythonPath(this._info.environment.path),
                     JUPYTER_TOKEN: this._info.token,
                     JUPYTER_CONFIG_DIR: process.env.JLAB_DESKTOP_CONFIG_DIR || app.getPath('userData')


mbektas (approved):
Thanks, LGTM! I don't think we need to filter out any env variables at this time.

PR #377, opened 2021-12-20
Upgrade to Electron 15

krassowski:
Looks like this one will be easy. Electron 15 blog post: https://www.electronjs.org/blog/electron-15-0
We are getting straight to 15.3.4.

Full changelog for 15.0.0: https://www.electronjs.org/releases/stable?page=3#15.0.0

Stack:
- Chromium 93 → 94
- Node v14.17.0 → v16.5.0
- V8: v9.3 → [v9.4](https://v8.dev/blog/v8-release-94)


mbektas (approved):
Great. Thanks!

mbektas:
I am getting kernel related errors once I install and test on my Mac. Released version 3.2.5-2 works fine on the same machine.

electron 15 problem

krassowski:
I can reproduce `Error during WebSocket handshake: Unexpected response code 403`. Nothing I tried so far helped:
- `nativeWindowOpen: false` in `BrowserWindow` constructor
- `--ServerApp.allow_remote_access=True`
- `--log_level=DEBUG` to check if the message gets received by the server and if there is more logs (it does not seem to be)
- downgrading Electron to 15.0.0
- setting custom CORS

I will appreciate any suggestions on how to tackle this one.

mbektas:
@krassowski can you rebase your branch so we can get test binaries with changes in https://github.com/jupyterlab/jupyterlab-desktop/pull/387?

krassowski:
Done

mbektas:
Thanks @krassowski . I tested on Mac and noticing two issues:
- PDF documents are not opening anymore
- After switching to Classic notebooks and clicking on a notebook opens a blank window
could you look into these please?

mbektas:
> Thanks @krassowski . I tested on Mac and noticing two issues:
> 
> * PDF documents are not opening anymore
> * After switching to Classic notebooks and clicking on a notebook opens a blank window
>   could you look into these please?

just checked the same with https://github.com/jupyterlab/jupyterlab-desktop/pull/387 and both work fine with those changes. so that makes me think these are failing maybe because of additional security measures in electron 15?

mbektas:
@krassowski `atob` change was necessary to get PDFs to open properly. PDF content is returned in base64 format from contents API but it contains whitespaces. This base64 content is converted to binary blob to render. atob function is used in the process. External browser's implementation of atob permits whitespaces. Since Desktop app creates BrowserWindow [without context isolation](https://github.com/jupyterlab/jupyterlab-desktop/blob/master/src/main/sessions.ts#L406), atob from node.js is exposed to browser window. atob of node.js does not permit whitespaces and causes PDF blob generation to fail. with my changes I am introducing a modified version of atob of node.js which allows whitespaces. Once we turn on contextIsolation, this patch wouldn't be needed anymore.

src/main/main.ts
```diff
@@ -190,10 +190,13 @@ app.on('ready', () => {
 });
 
 app.on("web-contents-created", (_event: any, webContents: WebContents) => {
-    // Prevent navigation
+    // Prevent navigation to local links
```
mbektas, line 193: @krassowski do you have any concerns on these changes? any concerns navigating to external websites?

krassowski:
> atob function is used in the process. External browser's implementation of atob permits whitespaces. Since Desktop app creates BrowserWindow [without context isolation](https://github.com/jupyterlab/jupyterlab-desktop/blob/master/src/main/sessions.ts#L406), atob from node.js is exposed to browser window.

Was it a change in Electron 15 which lead to this?

src/main/main.ts
krassowski: Yes, I am concerned. Wouldn't this lead to a context leak? Why is this change needed, again?

mbektas:
> 

yes, atob issue turns out to be caused by Electron 15 upgrade.

src/main/main.ts
mbektas: during my testing, external domain urls were opened in popup windows and that is safe, so I thought we could enable it back. but if you think there are cases that require blocking external urls then I can update it to block. The original motivation for this change was the issue I mentioned in earlier comments: `After switching to Classic notebooks and clicking on a notebook opens a blank window`

krassowski:
How about https://github.com/krassowski/jupyterlab-desktop/pull/15 ?

mbektas (approved):

PR #387, opened 2022-01-07
load all assets from localhost to set origin properly

mbektas:
- Loads initial content from `http://localhost:port/desktop-app-assets/index.html` to be able to set window.origin properly
- Fixes issues related to Electron 15 upgrade

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -8,17 +8,14 @@ import * as yargs from 'yargs';
 import * as path from 'path';
 import * as fs from 'fs';
 import { randomBytes } from 'crypto';
-import { createServer } from 'net';
+import { AddressInfo, createServer } from 'net';
 
 import { appConfig } from './utils';
 
-async function getFreePort(startPort: number): Promise<number> {
+async function getFreePort(): Promise<number> {
     return new Promise<number>((resolve) => {
-        let nextPort = startPort;
 
         const getPort = () => {
-            let port = nextPort++;
-
             const server = createServer((socket) => {
                 socket.write('Echo server\r\n');
                 socket.pipe(socket);
@@ -28,12 +25,13 @@ async function getFreePort(startPort: number): Promise<number> {
                 getPort();
             });
             server.on('listening', function (e: any) {
+                const port = (server.address() as AddressInfo).port;
                 server.close();
                 
                 resolve(port);
             });
         
-            server.listen(port, '127.0.0.1');
+            server.listen(0, '127.0.0.1');
         };
 
         getPort();
@@ -42,7 +40,7 @@ async function getFreePort(startPort: number): Promise<number> {
 
 async function setAppConfig(): Promise<void> {
     return new Promise<void>((resolve) => {
-        getFreePort(8888).then((port) => {
+        getFreePort().then((port) => {
             appConfig.jlabPort = port;
             appConfig.token = randomBytes(24).toString('hex');
             resolve();
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -7,6 +7,48 @@ import log from 'electron-log';
 import * as yargs from 'yargs';
 import * as path from 'path';
 import * as fs from 'fs';
+import { randomBytes } from 'crypto';
+import { createServer } from 'net';
+
+import { appConfig } from './utils';
+
+async function getFreePort(startPort: number): Promise<number> {
+    return new Promise<number>((resolve) => {
+        let nextPort = startPort;
+
+        const getPort = () => {
+            let port = nextPort++;
+
+            const server = createServer((socket) => {
+                socket.write('Echo server\r\n');
+                socket.pipe(socket);
+            });
+            
+            server.on('error', function (e) {
+                getPort();
+            });
+            server.on('listening', function (e: any) {
+                server.close();
+                
+                resolve(port);
+            });
+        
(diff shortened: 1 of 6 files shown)


src/main/main.ts
```diff
@@ -7,6 +7,48 @@ import log from 'electron-log';
...
+                
+                resolve(port);
+            });
+        
+            server.listen(port, '127.0.0.1');
+        };
+
+        getPort();
+    });
+};
```
krassowski, line 41: The Node.js documentation says:

> If port is omitted or is 0, the operating system will assign an arbitrary unused port, which can be retrieved by using `server.address().port`

Could we delegate the responsibility of choosing the port to the operating system to avoid potential issues?

src/main/main.ts
```diff
@@ -7,6 +7,48 @@ import log from 'electron-log';
...
+            server.on('error', function (e) {
+                getPort();
+            });
+            server.on('listening', function (e: any) {
+                server.close();
+                
+                resolve(port);
+            });
+        
+            server.listen(port, '127.0.0.1');
```
krassowski, line 36: The Node.js documentation for [`server.listen`](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback) says:

> If port is omitted or is 0, the operating system will assign an arbitrary unused port, which can be retrieved by using `server.address().port`

Could we delegate the responsibility of choosing the port to the operating system to avoid potential issues?

src/main/sessions.ts
```diff
@@ -420,12 +422,118 @@ class JupyterLabSession {
...
+            if (req.url.startsWith(desktopAppAssetsPrefix)) {
+                let assetPath = req.url.substring(desktopAppAssetsPrefix.length + 1);
+                const qMark = assetPath.indexOf('?');
+                if (qMark !== -1) {
+                    assetPath = assetPath.substring(0, qMark);
+                }
+                const assetFilePath = path.join(appAssetsDir, assetPath);
+                const assetContent = fs.readFileSync(assetFilePath);
+                callback(assetContent);
+            } else {
```
krassowski, line 482: Could you elaborate on what is happening in the `else` clause? My current understanding is that the `if ()` clase is for hosting static assets and `else` assumes that if it is not a static asset, then this is a server request and modifies cookies and handles file upload, but it is not clear to me _why_ specific parts are needed.

Should we split up the handler code into separate methods to better signal the intent and have easier time in the future?

src/main/sessions.ts
```diff
@@ -420,12 +422,118 @@ class JupyterLabSession {
...
+            const desktopAppAssetsPrefix = `http://localhost:${appConfig.jlabPort}/${DESKTOP_APP_ASSETS_PATH}`;
+            const appAssetsDir = path.normalize(path.join(__dirname, '../../'));
+    
+            if (req.url.startsWith(desktopAppAssetsPrefix)) {
+                let assetPath = req.url.substring(desktopAppAssetsPrefix.length + 1);
+                const qMark = assetPath.indexOf('?');
+                if (qMark !== -1) {
+                    assetPath = assetPath.substring(0, qMark);
+                }
+                const assetFilePath = path.join(appAssetsDir, assetPath);
```
krassowski, line 479: What if assets path includes `..`? That could potentially escalate access to filesystem. I think that we need to be very stringent here and check that the new `assetFilePath` is a child of `appAssetsDir`.

krassowski: We could also hard-code a list of static files that the client is allowed to access.

krassowski (commented):
Looks super exciting, looking forward to having this in!

While I see that this is still draft, I left some general comments to remember about these aspects later on.

src/main/main.ts
mbektas: good point. I will push an update.

src/main/sessions.ts
mbektas: yes, you are right. since interceptBufferProtocol doesn't provide a default callback handler, we are having to make the requests to JupyterLab server in the `else` section. During my experiments, I found out that cookies were not properly set for some of the requests (websocket handshake?). So, to be on the safe side, I am setting them with both `this._window.webContents.session.cookies.set(cookieObject);` and `requestHeaders['Cookie'] = Array.from(cookies.values()).join('; ');` for every request. I am not sure at the moment how we can optimize this.

src/main/sessions.ts
mbektas: I am not sure if asset path can include `..`. we are calculating it from request url. but still, It makes sense to  check if assetFilePath is a child of appAssetsDir, good idea.

src/main/utils.ts
```diff
@@ -29,6 +29,18 @@ import * as path from 'path';
...
+interface IAppConfiguration {
+    jlabPort: number;
+    token: string;
+};
+
+export
+const appConfig: IAppConfiguration = {
+    jlabPort: 8888,
+    token: 'jlab-token'
+};
```
krassowski, line 42: I anticipate that a hard-coded global will cause issues once we will go for full front/back separation. But I am happy to address it myself later.

krassowski (approved):
Thanks!

src/main/utils.ts
mbektas: yes, we can definitely improve this when we support remote backends etc. by the way, these are just default values and token is generated on the fly before server launch.

Issue #401, opened 2022-02-07
Jupyter Server Initialization Failed


Filthydude:
I don't understand what is going on. I have reinstalled jupiter eigth times and it still dose not work

[main.log](https://github.com/jupyterlab/jupyterlab-desktop/files/8014942/main.log)

connorferster:
I also have this problem although my log reads differently than @Filthydude: 

```
jlab --verbose 4

[3788:0207/200054.370:ERROR:cache_util_win.cc(20)] Unable to move the cache: Access is denied. (0x5)
[3788:0207/200054.371:ERROR:cache_util.cc(144)] Unable to move cache folder C:\Users\<me>\AppData\Roaming\jupyterlab-desktop\GPUCache to C:\Users\<me>\AppData\Roaming\jupyterlab-desktop\old_GPUCache_000
[3788:0207/200054.374:ERROR:disk_cache.cc(184)] Unable to create cache
[3788:0207/200054.374:ERROR:shader_disk_cache.cc(606)] Shader Cache Creation failed: -2
```

Installed JupyterLab 3.2.5-2 on Windows 10 Pro, 64-bit. Installed for "All Users" into default directories as follows:
* `C:\JupyterLab`
*  `C:\JupyterLab\resources\jlab_server`

Noticing that it was unable to "move the cache folder", I ran JupyterLab from the explorer "As Administrator..." to see if that would make a difference but it did not.

MichaDit:
Maybe related to #388.
Workaround: https://github.com/jupyterlab/jupyterlab-desktop/issues/388#issuecomment-1012115917

Issue #388, opened 2022-01-07
JupyterLab on Windows does not properly activate conda environment


ukoethe:
This might be related to #381 and #386.

After installation on Windows, I got the "Server Initialization Error" reported in #381 with same log message. I was able to work around the problem by starting jlab from the command line after **explicit activation of the Python environment**:
```cmd
C:\Users\ukoethe> C:\JupyterLab\resources\jlab_server\condabin\activate
(base) C:\Users\ukoethe> jlab .
```
The same was necessary in order to use a custom Python environment. Merely adding the path to the Python environment as described [here](https://github.com/jupyterlab/jupyterlab-desktop/blob/master/user-guide.md#changing-the-python-environment-used-by-jupyterlab-desktop) is **not sufficient**! Once again, the environment must first be activated, e.g.:
```cmd
C:\Users\ukoethe> activate my_env   # assuming conda is in the PATH
(my_env) C:\Users\ukoethe> jlab .
```
On Windows, the `activate` command adds about 7 directories to the `PATH` variable, and jupyter will not work properly without these settings. The workaround suggested in https://github.com/jupyterlab/jupyterlab-desktop/issues/381#issuecomment-1006790158 (copy some DLLs to other directories) is **not** the correct solution. 

I suggest to add the appropriate `activate` call at the beginning of `JupyterLab.exe`.

On a related note, in some cases the JupyterLab GUI opened, but was unable to start the kernel. After some time-out period, the status bar displayed `Python 3 | Disconnected`. Then, running a cell had no effect except for advancing to the next cell, as described in #386. Once again, activating the environment properly before running JupyterLab fixed this problem.

krassowski:
Thank you @ukoethe this is a very valuable observation. I really appreciate you taking time to experiment and figure out the problem and the solution.

aravindh-krishnamoorthy:
I can confirm that the solution in the OP works.

Installation version: v3.2.5-2 installed via the Windows Installer
Options: Just me + Installation directory: C:\Jupyterlab

Initially, I could not start JupyterLab Desktop due to #381. Then, I followed the instructions in the OP (I have no custom Python installation):

```
C:\Users\ukoethe> C:\JupyterLab\resources\jlab_server\condabin\activate
(base) C:\Users\ukoethe> jlab .

```

Now, JupyterLab Desktop opens.

yellowincarnate:
[...]

Issue #386, opened 2022-01-06
No output from running cells after installation on Windows


astroStudent123:
## Description

I installed on Windows using the instructions in the README file. Everything appears fine, except running cells produces no output; e.g. if I open a new notebook, create a code cell with the code print('here'), and click the "run the selected cells and advance" button, the curser advances but no output is produced. Re-installing doesn't fix this. 

## Reproduce

1. Download .exe file from https://github.com/jupyterlab/jupyterlab-desktop/blob/master/README.md
2. Install following setup prompts
3.  Open new notebook from the launcher
4.  Type any command that should produce output (e.g. a print statement)
5.  Click run cell 

## Expected behavior

Output should be produced but it isn't. The curser moves to the next cell but no output is produced. 

## Context

- Operating System and version: Windows 10 pro
- JupyterLab-Desktop version: Version 3.2.5-2
- Installer: .exe 

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the terminal inside of JupyterLab-Desktop here.
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.   

Try the new cross-platform PowerShell https://aka.ms/pscore6
PS C:\Users\Lea> jupyter troubleshoot                    
jupyter : The term 'jupyter' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the 
name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ jupyter troubleshoot
+ ~~~~~~~
    + CategoryInfo          : ObjectNotFound: (jupyter:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
 
PS C:\Users\Lea> 
![image](https://user-images.githubusercontent.com/90289357/148429523-ab15acf8-8077-4d80-a447-85d69f97e569.png)

Hmm... the terminal doesn't recognize this command. 

You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Browser Output (required for interface issues after loading)</summary>

<pre>
Paste the output from the DevTools JavaScript console here, if applicable.
 
![image](https://user-images.githubusercontent.com/90289357/148435364-afa5db72-a890-4e47-a10d-8808bebaf1f4.png)

</pre>
</details>

<details><summary>Logs (required for issues with loading, server and startup)</summary>

<pre>
Paste the logs from the mian.log file here.

[...]

Issue #381, opened 2021-12-23
Jupyter Server Initialization Failed


monomycelium:
I downloaded JupyterLab by following the recommended options and steps. When I launched the application, this showed:

![Issue](https://user-images.githubusercontent.com/95113443/147246602-e2614f50-72be-409b-89c4-306e34e5eaf6.png)

I tried launching the application again but the issue persisted. I uninstalled both the application and the server application and installed it again, but the same thing happened.

More information:
- Operating System and version: Windows 10 Pro (21H2) (Build 19044.1415)
- JupyterLab-Desktop version: 3.2.5-2
- Installer: .exe

Here is the log:
[main.log](https://github.com/jupyterlab/jupyterlab-desktop/files/7769521/main.log)

I hope this issue can be resolved soon as I cannot wait to try out this amazing application!

krassowski:
Thank you for reporting this issue. The log file includes the following:

```python
[2021-12-23 20:46:54.243] [info]  Jupyter Server initialization message: Traceback (most recent call last):
  File "C:\JupyterLab\resources\jlab_server\lib\runpy.py", line 185, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "C:\JupyterLab\resources\jlab_server\lib\runpy.py", line 144, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "C:\JupyterLab\resources\jlab_server\lib\runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "C:\JupyterLab\resources\jlab_server\lib\site-packages\jupyterlab\__init__.py", line 7, in <module>
    from .labapp import LabApp
  File "C:\JupyterLab\resources\jlab_server\lib\site-packages\jupyterlab\labapp.py", line 15, in <module>
    from jupyter_server.serverapp import flags
  File "C:\JupyterLab\resources\jlab_server\lib\site-packages\jupyter_server\serverapp.py", line 42, in <module>
    from jupyter_server.utils import run_sync_in_loop, urljoin, pathname2url
  File "C:\JupyterLab\resources\jlab_server\lib\site-packages\jupyter_server\utils.py", line 23, in <module>
    from tornado.httpclient import AsyncHTTPClient
  File "C:\JupyterLab\resources\jlab_server\lib\site-packages\tornado\httpclient.py", line 42, in <module>
    import ssl
[...]

PR #404, opened 2022-02-08
launch JupyterLab server via a script to activate the environment properly

mbektas:
- creates a temporary script file that activates the conda / venv environment properly and launches JupyterLab server
- follows @bollwyvl 's suggestion in https://github.com/jupyterlab/jupyterlab-desktop/pull/317#issuecomment-962683018
- tested on Windows, macOS, Ubuntu and Fedora
fixes https://github.com/jupyterlab/jupyterlab-desktop/issues/401 https://github.com/jupyterlab/jupyterlab-desktop/issues/388 https://github.com/jupyterlab/jupyterlab-desktop/issues/381 https://github.com/jupyterlab/jupyterlab-desktop/issues/386

### DIFF
--- a/src/main/registry.ts
+++ b/src/main/registry.ts
@@ -277,7 +277,7 @@ export class Registry implements IRegistry {
         const envName = `${envInfo.type}: ${envInfo.name}`;
 
         return {
-            type: IEnvironmentType.PATH,
+            type: envInfo.type === 'conda' ? IEnvironmentType.CondaEnv : IEnvironmentType.VirtualEnv,
             name: envName,
             path: pythonPath,
             versions: { 'python': pythonVersion, 'jupyterlab': jlabVersion },
--- a/src/main/registry.ts
+++ b/src/main/registry.ts
@@ -316,7 +316,7 @@ export class Registry implements IRegistry {
         if (platform === 'win32') {
             path_env = `${envPath};${envPath}\\Library\\mingw-w64\\bin;${envPath}\\Library\\usr\\bin;${envPath}\\Library\\bin;${envPath}\\Scripts;${envPath}\\bin;${process.env['PATH']}`;
         } else {
-            path_env = `${envPath}:${process.env['PATH']}`;
+            path_env = `${envPath}:${envPath}/bin:${process.env['PATH']}`;
         }
         
         return path_env;
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -63,7 +63,7 @@ function createLaunchScript(environment: IPythonEnvironment): string {
     if (isWin) {
         script = `
             CALL ${envPath}\\condabin\\activate
-            START ${launchCmd}`;
+            CALL ${launchCmd}`;
     } else {
         script = `
             ${envPath}/bin/activate
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -163,6 +163,7 @@ class JupyterServer {
 
         this._startServer = new Promise<JupyterServer.IInfo>((resolve, reject) => {
             const home = process.env.JLAB_DESKTOP_HOME || app.getPath('home');
+            const isWin = process.platform === 'win32';
             const pythonPath = this._info.environment.path;
             if (!fs.existsSync(pythonPath)) {
                 dialog.showMessageBox({message: `Environment not found at: ${pythonPath}`, type: 'error' });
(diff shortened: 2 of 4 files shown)


src/main/registry.ts
```diff
@@ -316,7 +316,7 @@ export class Registry implements IRegistry {
         if (platform === 'win32') {
             path_env = `${envPath};${envPath}\\Library\\mingw-w64\\bin;${envPath}\\Library\\usr\\bin;${envPath}\\Library\\bin;${envPath}\\Scripts;${envPath}\\bin;${process.env['PATH']}`;
         } else {
-            path_env = `${envPath}:${process.env['PATH']}`;
+            path_env = `${envPath}:${envPath}/bin:${process.env['PATH']}`;
```
bollwyvl, line 319: if possible, avoid doing _any_ `PATH` hacking, if it can possibly be avoided.

src/main/server.ts
```diff
@@ -68,21 +116,9 @@ class JupyterServer {
...
-            // removed once jupyter_server requires traitlets>5.0
-            this._nbServer = execFile(this._info.environment.path, [
-                '-m', 'jupyterlab',
-                '--no-browser',
-                // do not use any config file
-                '--JupyterApp.config_file_name=""',
-                `--ServerApp.port=${appConfig.jlabPort}`,
-                // use our token rather than any pre-configured password
-                '--ServerApp.password=""',
-                '--ServerApp.allow_origin="*"',
```
bollwyvl, line 82: eek!

src/main/server.ts
```diff
@@ -68,21 +116,9 @@ class JupyterServer {
             this._info.url = `http://localhost:${appConfig.jlabPort}`;
             this._info.token = appConfig.token;
 
-            // note: traitlets<5.0 require fully specified arguments to
-            // be followed by equals sign without a space; this can be
-            // removed once jupyter_server requires traitlets>5.0
-            this._nbServer = execFile(this._info.environment.path, [
```
bollwyvl, line 74: this well-quoted approach is probably still more robust against path shenanigans. 

If you know where the conda shell is, at least _it_ will run. You can couple that with the `run` subcommand. So it coudl actually be:

```ts
[
  pathToCondaBin,
  'run', 
  '--path', pathToBaseEnv,
  '--no-capture-output',
  '--cwd', pathToStartIn,
  'jupyter-lab',
```

src/main/server.ts
```diff
@@ -68,21 +116,9 @@ class JupyterServer {
...
-                '-m', 'jupyterlab',
```
bollwyvl, line 75: `-m` is not a great daily driver, as it effectively adds cwd to `PYTHONPATH`... it's very easy for local paths to clobber Important Python Stuff.

src/main/server.ts
krassowski: Nice. Still need some if/else as not every environment uses `conda`.

src/main/server.ts
```diff
@@ -29,9 +29,57 @@ import {
...
+    const pythonPath = environment.path;
+    let envPath = path.dirname(pythonPath);
+    if (!isWin) {
+        envPath = path.normalize(path.join(envPath, '../'));
+    }
+
+    // note: traitlets<5.0 require fully specified arguments to
+    // be followed by equals sign without a space; this can be
+    // removed once jupyter_server requires traitlets>5.0
+    const launchCmd = `python -m jupyterlab --no-browser --JupyterApp.config_file_name="" --ServerApp.port=${appConfig.jlabPort} --ServerApp.password="" --ServerApp.allow_origin="*" --ContentsManager.allow_hidden=True`;
```
krassowski, line 59: One neat way to preserve the comments on why we need particular options (which will otherwise become forgotten) 
would be to have the arguments as a list of strings and concatenate once ready.

src/main/server.ts
```diff
@@ -29,9 +29,57 @@ import {
...
+
+    let script: string;
+
+    if (isWin) {
+        script = `
+            CALL ${envPath}\\condabin\\activate
```
krassowski, line 65: Will this work for non-conda environments on Windows?

mbektas: good point, it won't as is. I am still testing these changes on Windows and will look into this as well.

src/main/server.ts
mbektas: @bollwyvl calling `jupyter lab` is failing on Windows with error `Error executing Jupyter command 'lab': [WinError 5] Access is denied `. that's why we have been using `python -m`

src/main/registry.ts
mbektas: the path here is not used when activating the environment anymore.

src/main/server.ts
mbektas: I chose to go with `source activate` since it works both for conda and venv. any concerns about that? does `conda run` have any advantage?

PR #415, opened 2022-03-03
fix python path used on createLaunchScript

anaximeno:
I created this pull request as a solution to issue #381 Jupyter Server Initialization Failed.

After analyzing the logs and the commands used on the processes called to execute the program, I figured that you were using the python of the console instead of the one on the environment, which wasn't allowing the app to launch properly.

To prevent the usage of the wrong python path I used the one that was stored on the constant **pythonPath**.

I also figured that it is being used `source env_path\bin\activate` to activate all types of environments on Linux desktops, which, if I am not wrong, can be the reason for the problem, since newer versions of Anaconda (and Miniconda) are not using the *activate* and *deactivate* binaries by default. I didn't change anything related to this problem (I just left a note comment) because the first solution was able to solve the main problem, however, I believe this needs to be revised.

### DIFF
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -61,7 +61,8 @@ function createLaunchScript(environment: IPythonEnvironment): string {
     // be followed by equals sign without a space; this can be
     // removed once jupyter_server requires traitlets>5.0
     const launchCmd = [
-        'python', '-m', 'jupyterlab',
+        // use the python path of the environment
+        `${pythonPath}`, '-m', 'jupyterlab',
         '--no-browser',
         // do not use any config file
         '--JupyterApp.config_file_name=""',
@@ -86,6 +87,7 @@ function createLaunchScript(environment: IPythonEnvironment): string {
                 CALL ${launchCmd}`;
         }
     } else {
+        // note: this will not work with newer versions of Anaconda
         script = `
             source ${envPath}/bin/activate
             ${launchCmd}`;


mbektas:
@anaximeno thanks for looking into this issue and the PR. we call `source ${envPath}/bin/activate` first, and that is supposed to update the PATH env variable so that when `python` is called it would run the python that is in the activated environment. I am not clear why your change is still required here.

have you tested the app with your changes to confirm that it resolves the issues you had? You can get the installer built by your PR from: https://github.com/jupyterlab/jupyterlab-desktop/actions/runs/1925757669

anaximeno:
Hi, you are welcome, and thanks for replying.

Yeah, I noticed that call before, but it seems not to be working for newer versions of Anaconda (>= 4.6).
To properly activate the environment named, by example **MyEnv**, on newer versions you would have to first activate the base environment using `source ${pathToTheBaseCondaEnv}\bin\activate` and then activate the environment using 
`conda activate ${MyEnv}`.

This happens because now the binaries to **activate** and **deactivate** can only be found on the **base environment**, so when the original command `source ${envPath}/bin/activate` was being called for environments that aren't the base, the environment wasn't being activated, and the jupyterlab wasn't being launched, as a result, I was receiving an error message whenever I attempted to open the app on that environments.

After changing the path used to call python to directly call the one from the environment, I did not get this error anymore.

mbektas:
@anaximeno `source bin/activate` is activating the environment properly for me, I am not needing to call `conda activate` after that. Below is my terminal output from Ubuntu. Can you try the same commands below and share your output?

```
mbektas@ubuntu:/opt/JupyterLab/resources/jlab_server$ source bin/activate
(base) mbektas@ubuntu:/opt/JupyterLab/resources/jlab_server$ conda --version
conda 4.11.0
(base) mbektas@ubuntu:/opt/JupyterLab/resources/jlab_server$ conda list jupyterlab
# packages in environment at /opt/JupyterLab/resources/jlab_server:
#
# Name                    Version                   Build  Channel
jupyterlab                3.3.0              pyhd8ed1ab_0    conda-forge
jupyterlab_pygments       0.1.2              pyh9f0ad1d_0    conda-forge
jupyterlab_server         2.10.3             pyhd8ed1ab_0    conda-forge
jupyterlab_widgets        1.0.2              pyhd8ed1ab_0    conda-forge
(base) mbektas@ubuntu:/opt/JupyterLab/resources/jlab_server$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal
```

anaximeno:
@mbektas for environments that are non base I would have trouble trying to activate it, since there's no activate script on the bin directory:

```bash
ax@mint:~$ cd anaconda3/
ax@mint:~/anaconda3$ source bin/activate 
(base) ax@mint:~/anaconda3$ ls envs/
Audio  Flask  ML  NLP  Shazam
(base) ax@mint:~/anaconda3$ ls envs/NLP/       # [comment] This is the env I am trying to activate
bin              conda-meta  etc      lib      licensing  mkspecs      plugins  resources  ssl           var                          x86_64-conda-linux-gnu
compiler_compat  doc         include  libexec  man        phrasebooks  qml      share      translations  x86_64-conda_cos6-linux-gnu
(base) ax@mint:~/anaconda3$ source envs/NLP/bin/activate     # [comment] That's the command you are using on the program to activate
bash: envs/NLP/bin/activate: Ficheiro ou pasta inexistente
(base) ax@mint:~/anaconda3$ ls envs/NLP/bin/ | grep activate   # [comment] There's no output, so, there's no activate on the dir envs/NLP/bin/
```
### To activate non base environments I could:

Using the **conda** command:

```bash
(base) ax@mint:~/anaconda3$ conda activate NLP
(NLP) ax@mint:~/anaconda3$ 
```

Now using the **bin/activate** script of the **base** environment and the **name** of the **environment** I want to activate:
```bash
(base) ax@mint:~/anaconda3$ source bin/activate NLP
(NLP) ax@mint:~/anaconda3$ 
```
There's also another option **similar to the previous**, but using the **path** to the environment:

```
(base) ax@mint:~/anaconda3$ source bin/activate envs/NLP/
(NLP) ax@mint:~/anaconda3$ 
```

mbektas:
@anaximeno thanks for the clarification. I wasn't aware that non-base conda envs didn't have the `bin/activate` script. In order to properly activate the non-base environments, looks like we need to locate the base environment and use the conda binaries from there.

Are you still needing the changes in this PR to be able to activate the `jlab_server` conda environment which is bundled with JupyterLab Desktop?

Also, since source bin/activate would fail for your non-base conda env, does this change in the PR work for your non-base env?

anaximeno:
@mbektas I opted to directly call the python on the environment I want to use, so, it wouldn't be necessary to activate the environment before executing the command. But, if you still want or need to activate the environment, it could be done by  activating your bundled (base) environment and then use the conda service on that environment to activate other non-base environments using the absolute path of that environments, as an example if I want to activate my environment named NLP, I could:

```bash
ax@mint:~$ source /opt/JupyterLab/resources/jlab_server/bin/activate 
(base) ax@mint:~$ conda activate /home/ax/anaconda3/envs/NLP
(NLP) ax@mint:~$ conda list jupyterlab
# packages in environment at /home/ax/anaconda3/envs/NLP:
#
# Name                    Version                   Build  Channel
jupyterlab                3.2.9              pyhd8ed1ab_0    conda-forge
jupyterlab-widgets        1.0.2                    pypi_0    pypi
jupyterlab_pygments       0.1.2              pyh9f0ad1d_0    conda-forge
jupyterlab_server         2.10.3             pyhd8ed1ab_0    conda-forge
(NLP) ax@mint:~$ 

```

Or even directly using the activate script of the bundled (base) environment:

```bash
ax@mint:~$ source /opt/JupyterLab/resources/jlab_server/bin/activate /home/ax/anaconda3/envs/NLP
(NLP) ax@mint:~$ conda list jupyterlab
# packages in environment at /home/ax/anaconda3/envs/NLP:
#
# Name                    Version                   Build  Channel
jupyterlab                3.2.9              pyhd8ed1ab_0    conda-forge
jupyterlab-widgets        1.0.2                    pypi_0    pypi
jupyterlab_pygments       0.1.2              pyh9f0ad1d_0    conda-forge
jupyterlab_server         2.10.3             pyhd8ed1ab_0    conda-forge
(NLP) ax@mint:~$ 
```

I will make some tests and analyze which option is the better one.

mbektas:
@anaximeno I implemented a solution to activate non-base conda environments in this PR: https://github.com/jupyterlab/jupyterlab-desktop/pull/425

could you test and confirm if it works for you? installers are at: https://github.com/jupyterlab/jupyterlab-desktop/actions/runs/1984671447

anaximeno:
@mbektas Your solution seems to work nicely on my computer.

mbektas:
fixed with https://github.com/jupyterlab/jupyterlab-desktop/pull/425, thanks @anaximeno!

PR #416, opened 2022-03-05
Pass explicit schema path to the JupyterLab server

krassowski:
https://github.com/jupyterlab/jupyterlab-desktop/pull/413#issuecomment-1057670346 highlighted an issue with schemas not being properly loaded (and hence toolbars not working) when upgrading to 3.3. It appears that the server was resolving the path to `json` schema files not to our assets, but to the assets in the selected user environment; if users had an old JupyterLab version installed, they would get old assets (missing required toolbar definitions, or absent altogether):

```
Failed to load toolbar items for factory Notebook from @jupyterlab/notebook-extension:panel Error: Schema not found: my-environment-of-choice-path/share/jupyter/lab/schemas/@jupyterlab/notebook-extension/panel.json
    at Function.create (browser.bundle.js:104932)
    at async SettingManager.fetch (browser.bundle.js:105995)
```

To reiterate, it was looking in `my-environment-of-choice-path` rather than in the path where JupyterLab Desktop is installed.

This PR solves the issue for now. It is not clear to me if this is the optimal approach.

### DIFF
--- a/src/main/server.ts
+++ b/src/main/server.ts
@@ -48,7 +48,7 @@ function createTempFile(fileName = 'temp', data = '', encoding: BufferEncoding =
     return tmpFilePath;
 }
 
-function createLaunchScript(environment: IPythonEnvironment): string {
+function createLaunchScript(environment: IPythonEnvironment, appPath: string): string {
     const platform = process.platform;
     const isWin = platform === 'win32';
     const pythonPath = environment.path;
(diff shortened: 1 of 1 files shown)


mbektas (approved):
Thanks @krassowski ! the schemasDir is not quite right for the release bundle but I will patch that. Otherwise the approach fixes the issue!

PR #420, opened 2022-03-09
use proper schemas dir when saving settings

### LINKED ISSUE #419: Failing to save Windows user settings with JupyterLab 3.3.0 upgrade
## Description

After JupyterLab 3.3.0 upgrade, user settings (including Python environment customization) are failing to save. Not an issue on Mac.

mbektas:
Uses schemas directory from bundled app directory. Fixes #419


mbektas:
Tested on Windows and Ubuntu. working fine!

PR #487, opened 2022-09-03
Bump eslint to 8.17

krassowski:
I intend to "no-enable-floating-promises" rule to tackle some of the issues seen in https://github.com/jupyterlab/jupyterlab-desktop/issues/438, this PR makes a first step by syncing the eslint version. It also resolves the warning that eslint is not compatible with the current typescript version.

PR #488, opened 2022-09-04
Improve error handling on frontend & fix some floating promises

krassowski:
Builds on top of #487, enabling no-floating-promises rule and handling errors on frontend; should lead to more informative user reports for frontend issues (although these are not as common as server issues).

Also ensures that root path is available when opening files.


mbektas (approved):
LGTM. Thanks!

mbektas:
@krassowski is this still in draft?

krassowski:
I have some local work on sending server errors to frontend and there is a conflict to resolve. If you are thinking about cutting a release I would **not** wait for me to finish it as I may not have enough time this weekend.

krassowski:
Closing as outdated.

Issue #438, opened 2022-04-07
Failed to find a compatible python environment at the configured path (3.3.2-2+)


sarkrui:
## Description

With the latest release (v3.3.2-2) of Jupyterlab Desktop, I can't use python env created by conda. It prompted with the following message. With v3.3.2-1, I didn't have any issue.
> failed to find a compatible python environment at the configured path "/users/p.xing/library/jupyterlab-desktop/jlab_server/bin/python". environment python package requirements are: jupyterlab >=3.1.0.

## Reproduce

- Install the latest v3.3.2-2
- activate python env with `conda activate jupyterlab`
- get python dir ` which python`
- Use the prompted path `/Users/myUserName/miniconda3/envs/jupyterlab/bin/python` as the Jupyter lab desktop custom env path
- failed to boot

## (pre) Solution

Using the v3.3.2-1 pkg installer, boot up with no issue.

ozogxyz:
I have the exact same problem on 3.3.4-2 in which I followed the exact steps while installing and trying to use my custom Conda environment.

ozogxyz:
I've found a workaround for this. I have copied the whole directory of my custom Jupyter environment (which includes Jupyter Lab 3.4.2 >= 3.1.0). Here are steps to reproduce:

The host machine is Monterey 12.2 with M1 Pro chip.

- Installed as usual, selected "Use the bundled Python environment"
- copied `~/miniforge3/envs/my_jupyter_env` to `~/Library/jupyterlab-desktop/`
- selected `~/Library/jupyterlab-desktop/my_jupyter_env/bin/python` as the custom Python environment

I created the `my_jupyter_env` with following packages:
- conda create -n jupyter python ipython jupyterlab nodejs nb_conda_kernels

Now Jupyter Lab Desktop's detects any other Conda environment with `ipykernel` installed in it. This way Python shows up as a native Apple process instead of Intel. 

<img width="928" alt="Screen Shot 2022-05-13 at 15 53 13" src="https://user-images.githubusercontent.com/50372354/168287322-5d166d0c-f7a6-4207-a551-1b5ae5ae6593.png">

Of course this is not a solution, but it seems the problem is still open since v3.3.2-2.

yangcht:
Could someone fix this, please?  
Having this issue since v3.3.4

USDSOM:
Issue still exists on v3.3.4-2
[...]

PR #489, opened 2022-09-04
Check if sub-environments directory exists before traversing it

krassowski:
Fix miniconda environments, see https://github.com/jupyterlab/jupyterlab-desktop/issues/438#issuecomment-1205087645. Briefly, miniconda does not set up `envs` as it does not support sub-environments. The common denominator of the reports was use of `miniconda` or `miniforge`.


mbektas (approved):
LGTM. Thanks!

PR #492, opened 2022-09-06
Add a global flag to test for execution within electron

fcollonval:
This adds a simple way for JupyterLab extensions to figure out if they are running within Electron app or not.

### DIFF
--- a/src/browser/index.tsx
+++ b/src/browser/index.tsx
@@ -44,4 +44,7 @@ function main(): void {
   );
 }
 
+// Flag to test for Electron integration
+window.isElectron = true;
+
 window.onload = main;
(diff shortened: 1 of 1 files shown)


src/browser/index.tsx
```diff
@@ -44,4 +44,7 @@ function main(): void {
   );
 }
```
mbektas, line 46: You can declare this new property to prevent build errors. 
```suggestion
declare global {
  // eslint-disable-next-line @typescript-eslint/naming-convention
  interface Window {
    isElectron?: boolean;
  }
}

```

fcollonval: Thanks for the suggestion

mbektas (approved):
LGTM. Thanks!

PR #500, opened 2022-10-20
remote JupyterLab server connection

mbektas:
This PR adds ability to connect to remote JupyterLab Servers as backend of the desktop application instead of launching a server locally.
- Supports connection to JupyterLab servers requiring authentication such as SSO
- Selected remote URL is saved as part of user settings and connection is restored at next launch

Here is how it works: Application first tries to connect to the remote server in a separate window and user enters credentials as needed. This populates session data and cookies are captured. `jupyter-config-data` is also retrieved from the connected JupyterLab server. Then JupyterLab Desktop is loaded in the main window using the same session and data retrieved.

Server Configuration Dialog (previously environment setting dialog)
server configuration dialog

Example connection to a remote binder instance
remote server connection

Example connection to JupyterHub
jupyter hub connection

### DIFF
--- a/scripts/buildutil.js
+++ b/scripts/buildutil.js
@@ -176,20 +176,6 @@ if (cli.flags.checkVersionMatch) {
     process.exit(1);
   }
 
-  // check JupyterLab versions in scripts
-  let searchString = `"appVersion": "${appVersion}",`;
-  if (
-    !searchTextInFile(
-      path.resolve(__dirname, `../src/browser/index.html`),
-      searchString
-    )
-  ) {
-    console.error(
-      `src/index.html doesn't contain correct Application version ${appVersion}`
-    );
-    process.exit(1);
-  }
-
   checkExtensionImports();
 
   console.log('JupyterLab version match satisfied!');
--- a/scripts/buildutil.js
+++ b/scripts/buildutil.js
@@ -37,17 +37,6 @@ const cli = meow(
   }
 );
 
-function searchTextInFile(filePath, text) {
-  try {
-    const fileContent = fs.readFileSync(filePath, 'utf8');
-    return fileContent.includes(text);
-  } catch (e) {
-    console.error('Error searching for file content', e);
-  }
-
-  return false;
-}
-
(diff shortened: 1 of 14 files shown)


mbektas:
@ellisonbg @krassowski this is still WIP and the code changes need refactoring but I would like to get your early feedback on the MVP feature set for remote server connection.

blink1073 (approved):
I'm not a JS guy anymore, but visually the changes look good to me.  Thanks @mbektas!

krassowski:
Quick feedback from trying it out today:
- Critical: the version shown in the help dialog now is referring to the version of jupyterlab-server but not the jupyterlab-desktop assets which are being served. This means that user is practically using 3.5 frontend but are shown 3.4 from the backend. If we release this like that it will be difficult to support.
- Important for UX `Server: local` is making working with multiple environments harder. I would love it to be `Server: local (your-environment)` until we get something dedicated for environments
- NTH: Server icon is a tad to big (I can take care of that)

mbektas:
Thanks @krassowski , I will revert the version to show desktop client version. Regarding your second point, I was originally labeling like you suggested but it felt really long for a status bar item. Example: `Server: local (conda: jlab_server)`. Dropping environment type (`conda:`) might make it better. Similarly, should we include any details on the label for remote server? Detailed information is still available for both on hover as tooltip.  Please let me know what you think.

krassowski:
> I will revert the version to show desktop client version

Maybe showing both is useful? Not sure.

> Example: Server: local (conda: jlab_server). Dropping environment type (conda:) might make it better. Similarly, should we include any details on the label for remote server? Detailed information is still available for both on hover as tooltip

Or should we drop `Server: ` since we already have a server icon and this part is constant? I think there is a tangible benefit from having than on screen rather than only in the tooltip/on click.

mbektas:
I can show both versions. I hope setting pageConfig["appVersion"] will not have any side effects. version can look like `Version 3.4.5-1 (desktop), 3.4.1 (server)`

Sure, removing `Server: ` sounds fine. I can add it to tooltip.

PR #508, opened 2022-11-04
Remote server connection UI enhancements

mbektas:
both versions on about dialog
version on about dialog

status bar item updates
server label and icon


krassowski (approved):
Looks great, thank you!

PR #510, opened 2022-11-05
Fix deprecation warnings on `publish` action

krassowski:
Seen on #508.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


mbektas (approved):
LGTM! thanks

PR #556, opened 2023-01-30
multiple window/session support

mbektas:
- Support for creating new sessions / windows using `New Window` menu item
- Support for restoring
- IPC event handling enhancements, proper event disconnect
- Synchronizing recent item list and python env list among windows


src/main/app.ts
```diff
@@ -43,6 +53,170 @@ interface IClearHistoryOptions {
...
+  }
+
+  private _getNewWindowRect(): IRect {
+    // cannot require the screen module until the app is ready, so require here
+    const { screen } = require('electron');
+    const primaryDisplay = screen.getPrimaryDisplay();
+    const {
+      width: screenWidth,
+      height: screenHeight
+    } = primaryDisplay.workAreaSize;
```
krassowski, line 107: So this does not work well for me with two-monitor setup. Say I have a setup like:

```
~~~~~
| 1  |  ~~~~~~~~~~~~
~~~~~   |          |
        |     2    |
        ~~~~~~~~~~~~
````

the window lands in the intersection of the bottom corner of 1 and top corner of 2. It is a bit impractical.

krassowski: [This comment](https://github.com/electron/electron/issues/3490) suggests using `.bounds`?

krassowski: There is also `.size` but neither `.bounds` nor `.workAreaSize` is well documented (at least not on https://www.electronjs.org/docs/latest/api/structures/display)

mbektas: can you test the latest? I updated with `bounds` origin.

mbektas: just tested and it worked better. pushed the change.

krassowski: This works well, cursor position is a neat heuristic. Thank you!

krassowski:
I kicked the macOS installer job (it got stalled).

mbektas:
> Preferences → Settings issue

I am not sure about this, maybe we can revisit later. there are other menu items with similar name clashes (macOS menu bar).

nick-youngblut:
Is there multi-window support for macs (apple silicon)? I only see the `Close Window` option (below), and I cannot find any other method of creating a new window. I'm using v4.2.5-1 of the desktop app.

Screenshot 2024-08-30 at 9 56 28 AM

I'm often working on multiple projects, so I need to jump between them. If the Jupyter desktop cannot handle multiple windows (projects), them I'm back to using the browser 😞

PR #561, opened 2023-02-03
remove logout and shutdown menu items

mbektas:
these menu items are causing issues with the desktop application. that's why hiding from JupyterLab File menu


krassowski:
Both entries would be better disabled by setting `quitButton: false` in PageConfig. Can we do that here?

mbektas:
> Both entries would be better disabled by setting `quitButton: false` in PageConfig. Can we do that here?

I wasn't aware of that config. I will use that. but I am closing this PR, will add to the other open one.

PR #562, opened 2023-02-03
Fixes before new UI release

mbektas:
- Sync session windows when user set Python environments are cleared
- Fix recent session collapse problem when a single recent item removed
- Log JupyterLab server launch parameters
- Ability to set log level from user settings, also configurable from Settings dialog
- Set default log level to warn instead of debug
- Remove Logout and Shutdown file menu items


krassowski (approved):
LGTM, thank you @mbektas!

PR #567, opened 2023-02-06
Release v3.6.1-1

mbektas:


user-guide.md
```diff
@@ -86,6 +106,45 @@ Make sure you installed JupyterLab Desktop following the steps outlined above in
   %pip install scikit-learn
   ```
 - In order to use the newly installed package you need to restart your active notebook's kernel or create a new notebook
+- If you install a new JupyterLab extension with UI components, then you will need to create a new session for changes to take effect
```
krassowski, line 109: I wonder if a restart button would be useful for this use-case. What I've been doing is opening the environment selection list and clicking enter to re-load server with the current environment, but with a dedicated button it could save me three click (and be obvious to every user).

user-guide.md
```diff
@@ -86,6 +106,45 @@ Make sure you installed JupyterLab Desktop following the steps outlined above in
...
+
+## Copying configuration from previous installation
+
+You can transfer settings from previous JupyterLab installations into JupyterLab Desktop by copying them over to the new configuration path `{jlab-desktop-user-data-dir}` by following these steps:
+
+1. Run `jupyter --paths` to determine where your `config` is stored.
+2. Find directory called `lab` in one of these paths.
+3. If there is a `user-settings` directory in `lab`, you can copy the `lab` directory over to the configuration path `{jlab-desktop-user-data-dir}` of JupyterLab Desktop to keep your old JupyterLab settings.
+
+**Warning**: If you copy over settings from an older major version of JupyterLab (e.g. 2.x) those might cause an error on startup.
```
krassowski, line 147:
```suggestion
**Warning**: If you copy over settings from an older major version of JupyterLab (e.g. 2.x) those might cause an error on startup. Similarly, if you copy settings from a newer (major or minor) version you may see errors.
```

In point releases we often add new settings so this also applies (e.g. 3.6 → 3.5 might error out if someone changed `displayMode`)

src/main/main.ts
```diff
@@ -12,6 +12,17 @@ import { SettingType, userSettings } from './config/settings';
 let jupyterApp: JupyterApplication;
 let fileToOpenInMainInstance = '';
 
+async function appReady(): Promise<boolean> {
```
krassowski, line 15: I am not sure if this is related but I do see a few seconds delay between when the window shows up and when it starts populating. The fact that splash screen animation is disabled exacerbates the lag feeling. Something to think about in the future releases.

krassowski (approved):
Thank you so much @mbektas!

user-guide.md
mbektas: maybe we can extend environment select popup into server info popup, it can have restart button. also features like copy server URL to clipboard. it could also just go into the app menu.

src/main/main.ts
mbektas: it is probably the environment discovery that is taking the time. launching without welcome page and opening a file at launch is heavy on the system. launching a server and discovering environments etc. this is something we can optimize I think. we can take a look in future releases.

PR #636, opened 2023-04-07
remove client app mode

mbektas:
## Reasons for removal

Client App mode:
- creates a lot of complexity to load client UI assets vs server supplied assets/data. requires intercepting network requests for solution.
- has unresolved issues such as PDF rendering
- much more prone to compatibility issues with the custom Python environments due to jupyterlab version differences
- complicates build and JupyterLab version updates
- requires prefetching pageConfig
- doesn't provide any noticeable performance gain
- increases bundled app size
- is untested for releases


krassowski (approved):
I was using it past few days, looks good to me. As long as we can customise the settings (like theme), add custom menus and custom UI elements (debatable, I know) I think whatever implementation is easiest to maintain is better.

PR #647, opened 2023-04-25
Add winget publish github action

brl0:
This adds a github action to publish the windows version to the winget-pkgs repo.  The winget-releaser github action requires a personal token as `WINGET_TOKEN`.  Details for configuration are here: https://github.com/vedantmgoyal2009/winget-releaser

For now, I have also submitted a PR to winget-pkgs to update to the current version, using the underlying tool utilized by this action.

This is my first attempt at configuring this as an action, so please let me know if there are any questions, concerns, or issues.

Thanks!

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,0 +1,16 @@
+name: Publish to WinGet
+on:
+  release:
+    types: [released]
+jobs:
+  publish:
+    # Action can only be run on windows
+    runs-on: windows-latest
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@v2
+        # https://github.com/vedantmgoyal2009/winget-releaser
+        with:
+          identifier: ProjectJupyter.JupyterLab
+          release-tag: ${{ steps.package-info.outputs.version}}
+          version: ${{ steps.package-info.outputs.version}}
+          token: ${{ secrets.WINGET_TOKEN }}
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -11,6 +11,4 @@ jobs:
         # https://github.com/vedantmgoyal2009/winget-releaser
         with:
           identifier: ProjectJupyter.JupyterLab
-          release-tag: ${{ steps.package-info.outputs.version}}
-          version: ${{ steps.package-info.outputs.version}}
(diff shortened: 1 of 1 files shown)


vedantmgoyal2009 (approved):
Before merging, there are 2 things that need to be done:

1. Please add a GitHub token with `public_repo` and `workflow` scopes as a repository secret and rename the secret name in the workflow.
2. Fork [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) under @jupyterlab.

mbektas:
It is not ideal for us to fork [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) under https://github.com/jupyterlab. we were expecting someone from community to update winget-pkgs repo as needed. It appears publishing winget package will require us to maintain it. any thoughts @krassowski?

vedantmgoyal2009:
@mbektas if that's the case, then any maintainer can fork winget-pkgs repo under their personal account and the workflow will be updated accordingly.

krassowski:
> It is not ideal for us to fork [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) under https://github.com/jupyterlab.

I think we can keep a fork under a bot account, the fork location appears customisable as per https://github.com/vedantmgoyal2009/winget-releaser#use-fork-under-which-user-fork-user.

> It appears publishing winget package will require us to maintain it. any thoughts @krassowski?

If this is running on CI with minimal maintianer intervention I would be fine to proceed. However, I am a bit concerned with using `vedantmgoyal2009/winget-releaser` action without hash pin - it is unofficial and not too widely used. I would recommend to pin to a hash if were to merge it. There are plans to upstream it as per https://github.com/microsoft/winget-pkgs/issues/500#issuecomment-1440568118 so it seems promising long-term.

There is another, simpler way to handle this without a CI action on our side nor a fork, by using a "vanity URL" as per https://github.com/microsoft/winget-pkgs/issues/500#issuecomment-755435018. We are already discussing a possibility of using vanity URLs in https://github.com/jupyterlab/jupyterlab/pull/14675#discussion_r1229607277 (CC @fcollonval, @paddymul) so maybe having a vanity URL per OS and one catch-all coud be the way to go.

mbektas:
@brl0 @vedantmgoyal9 thanks for the contribution, I will merge and wrap it up.

PR #712, opened 2023-10-02
show python requirements not found message and option to install

mbektas:
- Shows relevant error message when Python dependency packages such as `jupyterlab` are not found.
- Provides option to copy the command to clipboard and installing the missing packages for user
- Adds ability to open app logs from Settings dialog and progress views
- `copyable-span` web component to support copying to clipboard from web views
- Additional logging for shell executions and environment errors

python-env-requirements-message
show-logs

### DIFF
--- a/src/main/registry.ts
+++ b/src/main/registry.ts
@@ -66,7 +66,7 @@ export class Registry implements IRegistry, IDisposable {
         moduleName: 'jupyterlab',
         commands: ['--version'],
         versionRange: new semver.Range(`>=${MIN_JLAB_VERSION_REQUIRED}`),
-        pipCommand: `"jupyterlab>=${MIN_JLAB_VERSION_REQUIRED}"`
+        pipCommand: 'jupyterlab'
       }
     ];
 
--- a/src/main/registry.ts
+++ b/src/main/registry.ts
@@ -233,6 +233,7 @@ export class Registry implements IRegistry, IDisposable {
 
   private _resolveEnvironmentSync(pythonPath: string): IPythonEnvironment {
     if (!this._pathExistsSync(pythonPath)) {
+      log.error(`Python path "${pythonPath}" does not exist.`);
       throw {
         type: PythonEnvResolveErrorType.PathNotFound
       } as IPythonEnvResolveError;
@@ -253,6 +254,9 @@ export class Registry implements IRegistry, IDisposable {
     }
 
     if (!this._environmentSatisfiesRequirements(env, this._requirements)) {
+      log.error(
+        `Required Python packages not found in the selected environment. You can install using '${this.getRequirementsPipInstallCommand()}'.`
+      );
       throw {
         type: PythonEnvResolveErrorType.RequirementsNotSatisfied
       } as IPythonEnvResolveError;
--- a/src/main/registry.ts
+++ b/src/main/registry.ts
@@ -12,7 +12,9 @@ import {
   IDisposable,
   IEnvironmentType,
   IPythonEnvironment,
-  IVersionContainer
+  IPythonEnvResolveError,
+  IVersionContainer,
+  PythonEnvResolveErrorType
 } from './tokens';
 import {
   envPathForPythonPath,
@@ -45,6 +47,7 @@ export interface IRegistry {
   getCurrentPythonEnvironment: () => IPythonEnvironment;
   getAdditionalPathIncludesForPythonPath: (pythonPath: string) => string;
   getRequirements: () => Registry.IRequirement[];
+  getRequirementsPipInstallCommand: () => string;
   getEnvironmentInfo(pythonPath: string): Promise<IPythonEnvironment>;
   getRunningServerList(): Promise<string[]>;
   dispose(): Promise<void>;
@@ -53,17 +56,17 @@ export interface IRegistry {
 }
 
 export const SERVER_TOKEN_PREFIX = 'jlab:srvr:';
+const MIN_JLAB_VERSION_REQUIRED = '3.0.0';
 
 export class Registry implements IRegistry, IDisposable {
   constructor() {
-    const minJLabVersionRequired = '3.0.0';
-
     this._requirements = [
       {
         name: 'jupyterlab',
         moduleName: 'jupyterlab',
         commands: ['--version'],
-        versionRange: new semver.Range(`>=${minJLabVersionRequired}`)
+        versionRange: new semver.Range(`>=${MIN_JLAB_VERSION_REQUIRED}`),
+        pipCommand: `"jupyterlab>=${MIN_JLAB_VERSION_REQUIRED}"`
       }
     ];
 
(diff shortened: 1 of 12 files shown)


krassowski:
I started testing it, but did not get to it - I will report back tomorrow.

krassowski:
So when I tried to open an environment with an older version of JupyterLab:

```
jupyterlab             3.4.4
jupyterlab-server      2.15.0
```

I got this distressing screen:

image

which has no environment picker so I cannot change the environment (without restarting app). I was able to use the hamburger menu to go to settings and grab the logs using the new "Show logs" button.

```
source "/home/krassowski/.pyenv/versions/<sanitized-out>/bin/activate"
        python -m jupyterlab --no-browser --expose-app-in-browser --ServerApp.port=43547 --ServerApp.password="" --ServerApp.token="jlab:srvr:<sanitized out>" --LabApp.quit_button=False --JupyterApp.config_file_name="" --ContentsManager.allow_hidden=True
[2023-10-06 09:31:09.239] [debug] Server launch parameters:
  [script]: /tmp/jlab_desktopIaMVVT/launch.sh
  [options]: {<sanitized out>}
[2023-10-06 09:31:09.589] [error] Error: Reader received error. Reason: Error during socket reconnect: code = 1006, reason = ,,
[2023-10-06 09:31:09.589] [error] Connection closed by the server
[2023-10-06 09:31:09.589] [error] Connection unexpectedly disconnected
[2023-10-06 09:31:10.074] [error] Failed to start server Error: Jupyter Server process terminated before the initialization completed
    at ChildProcess.<anonymous> (/opt/JupyterLab/resources/app.asar/build/out/main/server.js:306:28)
    at ChildProcess.emit (node:events:513:28)
    at Process.onexit (node:internal/child_process:291:12)
[2023-10-06 09:31:10.359] [warn]  Showing error: [object Object]
[2023-10-06 09:31:16.357] [warn]  Showing error: [object Object]
```

mbektas:
@krassowski was this environment shown in your environment selection list on top right or did you browse the python path to select it? This failure is not due to version incompatibility, running the `python -m jupyterlab` somehow failed.

krassowski:
It's both. I do not have any suggestions in the environment list auto-populated because I use pyenv so I always have to browser for path to select it (hence I wish there was auto-detection for non-conda environments). But once I select it it stays in the environment selection list forever. It used to work previously, so I guess there may be something broken.
Re-iterating, running `python -m jupyterlab` with exact line as above from command line works fine.

krassowski:
I tried another environment which starts up fine with jupyterlab 3.6.3. My hypothesis was that it was the older JupyterLab version, so I downgraded. After downgrading the jupyterlab in the environment the picker tooltip still says that it sees 3.6.3. This is odd. How does it get jupyterlab version?

mbektas:
I tried in a conda environment with the same versions you posted above, and it worked just fine.

krassowski:
Ok, I know how to repro it reliably. It is a conflict of versions problem - it was happening when I stated `jlab` from my `jupyterlab` dev. Because of path overshadowing and not using path isolation it was pulling JupyterLab 4.0 Python executables, but then I selected an old environment to test it and it failed with the screen above.

So this is not a typical failure scenario, but it does show that this error screen needs work (in particular it would be lovely if it had the environment selector).

src/main/sessionwindow/sessionwindow.ts
```diff
@@ -712,40 +791,50 @@ export class SessionWindow implements IDisposable {
...
+          this._registry.addEnvironment(path);
+        } catch (error) {
+          let message = `Error! Python environment at '${path}' is not compatible.`;
+          if (
+            error?.type === PythonEnvResolveErrorType.RequirementsNotSatisfied
+          ) {
+            const requirementInstallCmd = encodeURIComponent(
+              this._registry.getRequirementsPipInstallCommand()
+            );
+            message = `Error! Required Python packages not found in the selected environment. You can install using <copyable-span label="${requirementInstallCmd}" copied="${requirementInstallCmd}"></copyable-span> command in the selected environment.`;
```
krassowski, line 804: Is it possible to show the detected version here>

src/main/registry.ts
```diff
@@ -53,17 +56,17 @@ export interface IRegistry {
...
-    const minJLabVersionRequired = '3.0.0';
-
     this._requirements = [
       {
         name: 'jupyterlab',
         moduleName: 'jupyterlab',
         commands: ['--version'],
-        versionRange: new semver.Range(`>=${minJLabVersionRequired}`)
+        versionRange: new semver.Range(`>=${MIN_JLAB_VERSION_REQUIRED}`),
+        pipCommand: 'jupyterlab'
```
krassowski, line 69: should it include `>=${MIN_JLAB_VERSION_REQUIRED}` as well?

src/main/registry.ts
```diff
@@ -446,6 +468,16 @@ export class Registry implements IRegistry, IDisposable {
     return this._requirements;
   }
 
+  getRequirementsPipInstallCommand(): string {
```
krassowski, line 471: I am afraid that suggesting `pip` install will lead to many problems if novice conda users end up executing this in their env. Worst case scenario is having two conflicting versions with assets from one and scripts form another which is quite common problem but hard to debug.

src/main/sessionwindow/sessionwindow.ts
mbektas: it is possible. it is easier to just log them though.

src/main/registry.ts
mbektas: it was causing installation to fail on Windows. I will take another look.

src/main/registry.ts
mbektas: is this an issue when an old jupyterlab is installed with `conda install`, then updated with `pip install`. I thought pip install was fine in a conda environment.

krassowski: It will only work when NOT in the base environment (ideally in isolated environment) and if `pip` was explicitly installed. Many users fall into this trap.

Reference: https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#using-pip-in-an-environment

krassowski: Of course if you install the same version in base and leaf env, one using pip and the other using conda, it does not matter. Up until the very moment one gets updated and then it breaks.

mbektas:
@krassowski could you take another look as I pushed some updates addressing your comments.

krassowski:
I think I am seeing an issue with `PYTHONPATH`. Previously I used to set it like this (simplifying) :

```
PYTHONPATH=/my/project/root/dir:$PYTHONPATH jlab
```

but this does not seem to work now? I am not 100% positive if this is due to this PR or because some setting got lost during uninstall/install sequence, but given that this PR touches python path logic it could be related.

This is kind of a deal breaker for me, but possibly not a common usage pattern. Note that configuring this in settings does not seem to work either. Is Python path overwritten at some point, or am I fooling myself?

I am going to try an older version now.

mbektas:
> I think I am seeing an issue with `PYTHONPATH`
I tested both from CLI like your example and also from the settings dialog and both worked fine. This PR doesn't make a change to environment variable setting.

krassowski:
Thank you for looking into this! Apologies for the delay with the actual review, will take another look soon.

mbektas:
> A spinner, or even better logs during live installation would be a great future enhancement.

there is actually a subtle [animation](https://github.com/jupyterlab/jupyterlab-desktop/blob/master/src/assets/progress-logo.svg) of the orange part of the logo, like heart beat. but probably too subtle to notice.

PR #721, opened 2023-10-26
Release v4.0.7-1

mbektas:


krassowski:
Looking at release notes, I think it might be worth adding that Electron 27 upgrade implies that "macOS 10.13 (High Sierra) and macOS 10.14 (Mojave)" are no longer supported.

PR #742, opened 2023-12-26
Python env management UI and env selector popup updates

mbektas:
- Introduces Python environment management UI to manage existing environments and create new ones
- Create a new Python env by using bundled installer
- Create a new Python env by using conda or pip and selecting package list
- Python `envs` install directory config to create new envs under
- Updates Python environment selection popup, adds filtering support and link to env management dialog

Python env selection popup
new env select popup

Python env selection popup filtered
filter

Manage Python envs UI, list of existing envs
manage env 1

Manage Python envs UI, env menu
manage env menu

Manage Python envs UI, create new using bundled installer
manage env 2

Manage Python envs UI, create new using conda or pip
manage env 3

Manage Python envs UI, settings (server default env moves from settings dialog to here)
manage env4


mbektas:
@krassowski code is still WIP but could you provide feedback on the design & features?

krassowski:
when I type in something which leads to no match, I cannot remove it (backspaces does not work because the input box becomes inactive):

filtering-all-out

IMO the restart and copy buttons are a bit confusing.
- they represent actions related to current session, but the "x" button nearby represents action on the selector popup
- the text says "restart session using a **different** Python env" but they allow to restart using the same env

Maybe these should be two lines, like:

```
--------------------------------------------------------------------------------
| Current environment: full path    [restart] [copy]                       [x] |
| [a subtle divider]                                                           |
| Restart session with a different Python environment                          |
| [ Type to filter Python paths ]                                              |
--------------------------------------------------------------------------------
```

mbektas:
@krassowski I updated the popup as below
Screenshot 2024-01-21 at 10 23 00 AM

PR #751, opened 2024-01-23
Snap package

dmitry-lyfar:
## Description
Adds snap package building and publishing. 

* The chosen confinement for the snap is 'classic' due to the issues that the 'strict' confinement has when a user works with files in $HOME. JupyterLab server creates temporary hidden files in the working directory. The strict confinement prohibits creating those in the home directory which results in a permission denied error when JLD attempts to (auto-)save.
* GitHub workflows extended with steps that builds, uploads and publishes snap. There are two publish steps to allow publishing edge snaps built with the latest changes automatically and publishing stable releases to the latest/stable for the tagged commits. Publishing requires:
  * [Registering](https://snapcraft.io/docs/registering-your-app-name) a snap name (e.g. jupyterlab-desktop) in Snap Store.
  * Snap Store credentials that can be obtained as:
```
$ snapcraft export-login --snaps=jupyterlab-desktop \
      --acls package_access,package_push,package_update,package_release \
      exported.txt
```

@mbektas I have not registered a snap name as it should probably be done under [Jupyter Project](https://snapcraft.io/publisher/projectjupyter) publisher. When registered, we can obtain the store credentials and have the publishing workflow working. Also, when done, we would need to apply for the classic snap approval as only the strict ones are allowed for automatic publishing (I can take care of that as well).


mbektas:
@dmitry-lyfar I tested these changes locally (on a Ubuntu aarch64) and ran into the problems below. do you know of solutions for these by changing the snap configuration? or do we have to make changes to app code?

1. double clicking the snap file didn't install. this seems to be related to classic confinement. I had to run this command on CLI to install: `sudo snap install ./JupyterLab.snap --classic --dangerous`
2. The userData and home directories used by the app changed to subfolders in `~/snap/jupyterlab-desktop/x1/`. this causes issues with restoring settings from previous debian installation. but more concerning issue is in `#3` below
3. Uninstalling the app also removed the settings and Python environment installation by the app because they got installed into paths in `#2` and those got deleted.

dmitry-lyfar:
Hi @mbektas,

> 1. double clicking the snap file didn't install. this seems to be related to classic confinement. I had to run this command on CLI to install: `sudo snap install ./JupyterLab.snap --classic --dangerous`

Yes, this is because snaps are not distributed as standalone archives. When we publish, a user would be able to do `snap install jupyterlab-desktop --classic` from the CLI or just click a button if they use a GUI Snap Store desktop app. No need to download anything.

> 2. The userData and home directories used by the app changed to subfolders in `~/snap/jupyterlab-desktop/x1/`. this causes issues with restoring settings from previous debian installation. but more concerning issue is in `#3` below
> 3. Uninstalling the app also removed the settings and Python environment installation by the app because they got installed into paths in `#2` and those got deleted.
Do you mean the bundled Python environment that is created and installed into `~/snap/jupyterlab-desktop/current/.config/`? 

By default, the data created within snap user directories is snapshotted and stored for, I believe, 30 days (see `snap saved` et al) if the snap is removed. The data is removed as well but can be recovered if snap is re-installed or via `snap restore`. If you prefer the bundled environment to be preserved even if JLD snap is removed, I can have a look into. But I suppose that the bundled environment is something that has a life-time of the application. If needed for longer, a user should point the application to the pre-existing environment?

PR #764, opened 2024-02-06
notification for bundled env updates, auto update bundled env

mbektas:
- On session window title bar and Python environment selection popup, show notification and action button to update bundled environment installation if a newer version is available to install.
- Environment update options on Settings dialog

bundled env update settings

bundled env update badge

bundled env update action


src/main/app.ts
```diff
@@ -773,6 +791,27 @@ export class JupyterApplication implements IApplication, IDisposable {
...
     );
 
+    this._evm.registerEventHandler(
+      EventTypeMain.UpdateBundledPythonEnv,
+      async event => {
+        const choice = dialog.showMessageBoxSync({
+          type: 'warning',
+          message: 'Update bundled environment',
+          detail:
+            'App will restart and the existing environment installation will be deleted before update. Would you like to continue?',
```
krassowski, line 801: Should there be a "the" before "update"?

src/main/dialog/themedwindow.ts
```diff
@@ -114,7 +121,7 @@ export class ThemedWindow {
           <div class="page-container">
             <jlab-dialog-titlebar id="title-bar" data-title="${
               this._window.title
-            }" class="${
+            }" data-closable="${this._closable.toString()}" class="${
```
krassowski, line 124: Being explicit with embedding in HTML makes it one less thing to worry about.

```suggestion
            }" data-closable="${this._closable ? 'true' : 'false'}" class="${
```

src/main/pythonenvselectpopup/pythonenvselectpopup.ts
```diff
@@ -172,6 +182,9 @@ export class PythonEnvironmentSelectPopup {
             <div class="toolbar-button" id="copy-button" onclick='handleCopySessionInfo();' title="Copy session info to clipboard">
               ${copyIconSrc}
             </div>
+            <div class="toolbar-button" id="update-bundled-env-button" onclick='handleUpdateBundledEnv();' title="Update available for Python environment. Click to install.">
```
krassowski, line 185: Should it be hidden by default?

```suggestion
            <div class="toolbar-button" id="update-bundled-env-button" onclick='handleUpdateBundledEnv();' style="display: none;" title="Update available for Python environment. Click to install.">
```

It is not clear for me if `onShowUpdateBundledEnvAction` will fire when disabled in options.

src/main/titlebarview/titlebar.html
```diff
@@ -157,6 +169,7 @@
         <div id="server-button">
           <div id="server-icon">
             <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"><!--! Font Awesome Pro 6.2.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2022 Fonticons, Inc. --><path d="M64 32C28.7 32 0 60.7 0 96v64c0 35.3 28.7 64 64 64H448c35.3 0 64-28.7 64-64V96c0-35.3-28.7-64-64-64H64zM344 152c-13.3 0-24-10.7-24-24s10.7-24 24-24s24 10.7 24 24s-10.7 24-24 24zm96-24c0 13.3-10.7 24-24 24s-24-10.7-24-24s10.7-24 24-24s24 10.7 24 24zM64 288c-35.3 0-64 28.7-64 64v64c0 35.3 28.7 64 64 64H448c35.3 0 64-28.7 64-64V352c0-35.3-28.7-64-64-64H64zM344 408c-13.3 0-24-10.7-24-24s10.7-24 24-24s24 10.7 24 24s-10.7 24-24 24zm104-24c0 13.3-10.7 24-24 24s-24-10.7-24-24s10.7-24 24-24s24 10.7 24 24z"/></svg>
+            <div id="server-notification-badge"></div>
```
krassowski, line 172: I guess the same suggestion here to make it display none by default?

krassowski (approved):
Great enhancement, thank you!

src/main/pythonenvselectpopup/pythonenvselectpopup.ts
mbektas: it is hidden by default already using CSS. you can see in [sessionwindow.ts](https://github.com/jupyterlab/jupyterlab-desktop/pull/764/files#diff-17b1a4383e0d332e3de36bff535ba7f3afb757e7dee8839882b3484c4aea4790) that it is fired only if enabled in options.

src/main/titlebarview/titlebar.html
mbektas: it is hidden by default already using CSS

PR #781, opened 2024-02-26
UI modes

mbektas:
- Adds UI mode switch to session menu. UI mode is saved per project (working directory).
- Zen Mode is the simplified Single document mode. It is an alternative to Classic Notebook view.
- Custom mode is the default mode that comes with JupyterLab and restores user's layout configuration. Other modes reset the layout based on the selection whenever a new session is created.

- [x] Introduce UI modes
- [ ] Expose UI modes in settings dialog and CLI
- [ ] Make Zen Mode default for opening single notebook files by double clicking or from CLI

cc. @bollwyvl I looked into supporting Jupyter Notebook mode but it was not feasible. It didn't seem to worth the effort. The Zen Mode introduced in this PR is a good alternative for Classic Notebook view fans.

ui-modes-2

### DIFF
--- a/src/main/labview/labview.ts
+++ b/src/main/labview/labview.ts
@@ -21,6 +21,7 @@ import { SessionWindow } from '../sessionwindow/sessionwindow';
 import {
   CtrlWBehavior,
   SettingType,
+  UIMode,
   userSettings,
   WorkspaceSettings
 } from '../config/settings';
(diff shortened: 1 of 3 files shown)


src/main/labview/labview.ts
```diff
@@ -222,6 +223,87 @@ export class LabView implements IDisposable {
...
+    await this._setUIMode(uiMode);
+  }
+
+  async _setUIMode(uiMode: UIMode) {
+    this._uiMode = uiMode;
+
+    await this._view.webContents.executeJavaScript(`
+    executeJSResult_currentLayout = {};
+    {
+      const lab = window.jupyterapp || window.jupyterlab;
```
krassowski, line 245: I wonder if we should make the code below into a proper Jupyter extension and make it expose a single command on the window, like `window.setJupyterLabMode('zen' | 'multi' | 'single')`. This would ensure that as we update versions the type validation catches any changes, and allow re-using it elsewhere. This could be an extension or set of extensions in this repo, or even plugins which might be disabled by default in jupyterlab repo.

My main motivation would be to ultimately revert the change which exposed `jupyterapp` on window e.g. to avoid users writing logic which assumes that it is there (it will not be available outside of Desktop).

krassowski:
Looks amazing btw!

src/main/labview/labview.ts
mbektas: that makes sense. I was also thinking about exposing desktop API endpoints (something like window.jlabDesktopApp.<methods>), so that extensions can use. I will create a separate issue for such refactoring.

Issue #795, opened 2024-03-12
Document TOC in Zen mode?


fperez:
First, huge thanks for the new release and the new UI modes! I discovered them yesterday on update before even seeing the blog post, and loved it!

I do have one request/suggestion: while I love Zen mode, one thing that unfortunately makes it a non-starter for me is the loss of the TOC. When I'm focused on a single notebook (the whole point of Zen mode), one of the most important/useful things for me to have is easy navigation throughout my document's structure, and not by scrolling. For that, the TOC is absolutely critical, and it is for me one of the most valuable features we have in Lab today. It helps me stay in the flow of the document by quickly going to the part of the analysis/narrative I need.

I suspect it's a bit tricky from a UI perspective to offer it in Zen mode if you've entirely suppressed the left side panel, but I still think it would be worth exploring.

Otherwise congrats on the release/blog post and thanks for the great work!  I am using it more and more regularly as my main Jupyter UI instead of the web app!

mbektas:
Thanks for the feedback @fperez ! I will add an option to enable TOC in Zen Mode. I thought of possibly enabling other widgets (status bar / file explorer (instead of TOC)) in Zen Mode but Single Document Mode looks better fit for those use cases.

fperez:
Agreed @mbektas - I could go either way on the status bar (I can see it being useful even in zen mode and not distracting most users). But I totally think the file panel should stay out: this is precisely a desktop app, where people have a "real" file manager lying around from their OS (and typically a better one than ours), so it makes sense to delegate that task to the Desktop OS, and let our tools focus on one task: Jupyter.

PR #798, opened 2024-03-16
show TOC in Zen mode option

mbektas:
Option to show Table of Contents panel in Zen Mode
toc-in-zen-mode

lab-toc-in-zen-mode

fixes https://github.com/jupyterlab/jupyterlab-desktop/issues/795

### DIFF
--- a/src/main/labview/labview.ts
+++ b/src/main/labview/labview.ts
@@ -282,9 +282,9 @@ export class LabView implements IDisposable {
 
     await this._view.webContents.executeJavaScript(`
     {
-      jlabDesktop_setUIMode('${this._uiMode}', ${userSettings.getValue(
-      SettingType.showTOCInZenMode
-    )});
+      jlabDesktop_setUIMode('${this._uiMode}', {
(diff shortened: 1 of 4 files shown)


src/main/labview/labview.ts
```diff
@@ -463,7 +465,7 @@ export class LabView implements IDisposable {
           });
         }
 
-        async function jlabDesktop_setUIMode(uiMode) {
+        async function jlabDesktop_setUIMode(uiMode, showTOCInZenMode) {
```
krassowski, line 468: I would make the second argument an object with options like `options: {showTOCInZenMode?: boolean}` to make it easy to add/remove arguments in the future, but this is not really a public API (I think?) so probably does not matter much.

fperez (approved):
I'm not competent enough in TypeScript to comment on the code, but the UX is exactly what I was looking for. Thanks so much for being responsive!

I think this will make Zen mode a real favorite for those looking to focus on a document-centric experience.  Great work, thanks again!

PR #800, opened 2024-03-17
add jlab CLI for snap

mbektas:


electron-builder-scripts/snap-hooks/configure
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -e
+
+APP_PATH=$SNAP/jupyterlab-desktop
+ln -s "$APP_PATH" /usr/bin/jlab
```
dmitry-lyfar, line 4: @mbektas I would suggest adding -f as in my case, I already had `/usr/bin/jlab` for some reason and configure hook failed (which makes the installation to roll back completely):

`ln -sf "$APP_PATH" /usr/bin/jlab`

PR #880, opened 2024-12-18
Update to JupyterLab v4.3.4

mbektas:
New JupyterLab release [v4.3.4](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.3.4) is available. Please review the lock file carefully..


krassowski:
Thank you for testing @ijstokes, I will merge this now!

I think some immediate action items would be:
- the jupyterlab update action is failing (https://github.com/jupyterlab/jupyterlab-desktop/actions/runs/17105910291), this is the latest PR which it generated but we are far into 4.4.x cycle 
- we need to update electron to a supported version
- we should handle the sandboxing issues which prevent running on recent Ubuntu and MacOS smoothly

Issue #618, opened 2023-03-15
Open links in the system's default browser


palpen:
At present, clicking on a link in JupyterLab desktop will open the link in a separate window in JupyterLab. It would be great if 
instead the links are opened in the system's default browser (i.e. Chrome, Safari, etc).

Version of the app I'm using: Version 3.6.1-3 (3.6.1-3)

butuzov:
is there any updates on this question? maybe setting that can be used?

krlng:
This bug is basically the reason I open my jupyter-notebooks in vscode now...

Nagidrop:
I would love to have an option to enable this. Current behaviour seems somewhat limited.

jimgoo:
+1, deal breaker for me

notluquis:
Still reproducible on current master (Electron 42, JupyterLab 4.6.1 build). Clicking a URL rendered in a notebook traceback opens it in an in-app Electron window instead of the system browser. The app frame itself correctly stays put, so it is only the target that is wrong.

Looking at why: there is no `setWindowOpenHandler` anywhere under `src/`, and `git log -S` shows there never was one, so `window.open` falls through to Electron's default new-window behaviour. Besides the UX, an uncontrolled in-app window loading an arbitrary URL is a surface worth closing.

#883 already implements the fix in the right place (`src/main/labview/labview.ts`) but has gone stale and now conflicts with master. Parking this for the milestone after 4.6.1-1 rather than rushing it into the release.

I confirmed this by grepping the source and by clicking a traceback link in a local 4.6.1 build. Work done with Claude Code.

drnw:
I have recently upgraded to Version 4.6.2-1 and observe the feature requested appears to be implemented. That is that a URL click opens in the default browser rather than in a Jupyter Desktop window. 

For my use case this is problematic:

1) User opens session with a remote JupyterLab service which has access protection from Cloudflare.
2) Cloudflare's sign in with Github page is rendered in a Jupyter Desktop window.
3) User clicks link in Jupyter Desktop window and Github page opens in default browser
4) Github authenticates in default browser not Jupyter Desktop

=>  User cannot access the remote JupyterLab in Jupyter Desktop

krassowski:
[...]

PR #883, opened 2025-01-17
Open external links in system browser

ozpau:
This solves #618.

### DIFF
--- a/src/main/labview/labview.ts
+++ b/src/main/labview/labview.ts
@@ -6,7 +6,8 @@ import {
   clipboard,
   dialog,
   Menu,
-  MenuItemConstructorOptions
+  MenuItemConstructorOptions,
+  shell
 } from 'electron';
 import log from 'electron-log';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


ozellpaukert:
Found a problem with this implementation. If you click "Open in Notebook", then it will open notebook in system browser instead of jupyterlab-desktop.

ozellpaukert:
While working on this I discovered another problem (that is also present in the upstream):
1. Opening jupyter notebook from jupyter lab creates a window that is not decorated like main labview
2. Opening external links in jupyter windows opened from jupyter labs doesn't use system browser
3. Opening jupyter lab from jupyter notebook creates an undecorated window

notluquis:
Confirmed the underlying issue (#618) is still present on current master (Electron 42, JupyterLab 4.6.1 build): a link in a notebook traceback opens in an in-app Electron window rather than the system browser, and there is no `setWindowOpenHandler` under `src/` at all.

This PR still points at the right file (`src/main/labview/labview.ts`), it has just drifted and now conflicts with master. Parking it for the milestone after 4.6.1-1 so it does not keep sitting.

Verified against a local 4.6.1 build; done with Claude Code.

krassowski:
Closing as superseded by https://github.com/jupyterlab/jupyterlab-desktop/commit/7f8f64ae796efae77d106612c95f0edd6c5dc273. Thank you for the PR!

Issue #890, opened 2025-04-09
Should we archive this project?


krassowski:
Should we archive `JupyterLab Desktop`? Making regular new releases there is imperative from the security perspective (it is shipping a whole browser, with all its zero days), and it is also pretty high labour as the releases actually needs to be manually tested on MacOS, Windows, and Linux. I was helping with testing on Linux but currently neither have access to MacOS, nor availability to test all three in my free time. @mbektas did the most thorough testing back in the day and was managing the releases, so could possibly share some notes and thoughts there.

We previously discussed this on triage meetings when triaging issues in this repo, and last week on jupyter frontends meeting.

jtpio:
It would be sad to have to archive the repo, because JupyterLab Desktop is a valuable piece in the Jupyter Frontends offering, and it was really received when first announced.

But the security risks and maintenance burden mentioned above are indeed likely enough to justify archiving the repo. Or at least for now.

Ideally, some folks would be able to pick up development instead, and maybe https://github.com/jupyter-governance/jupyter-foundation-governing-board/issues/1 will help with this hopefully.

mbektas:
I used to maintain this app in my personal time and unfortunately not finding that much free time to contribute anymore. I would suggest Jupyter org to consider investing in this app using the new funds (https://github.com/jupyter-governance/jupyter-foundation-governing-board/issues/1), instead of sunsetting.

1. I am guessing there are 100K+ active users of this product that would be disappointed to lose it. ([see these stats](https://somsubhra.github.io/github-release-stats/?username=jupyterlab&repository=jupyterlab-desktop&page=1&per_page=5))
2. This app lowers the barrier of entry to Jupyter eco-system (one click installation) and adds new features to JupyterLab web app (i.e. ability to work with different configurations in different directories, opening notebooks by double click and more).
3. Sunsetting this tool would lead more users to other notebook apps such as VS Code.
4. I expect the security risk surface area to be much smaller than assumed. Because the app's browser is only loading JupyterLab application and its extensions, in comparison to random websites a person browses on a regular browser.

krassowski:
> I expect the security risk surface area to be much smaller than assumed. Because the app's browser is only loading JupyterLab application and its extensions, in comparison to random websites a person browses on a regular browser.

[...]

PR #897, opened 2025-08-07
Add a warning about the current state of the app

krassowski:
Related to #890


RRosio (approved):
Thanks for doing this Mike, I think it is a good idea to add this warning meanwhile there is further discussion about the state of the project.

PR #910, opened 2025-08-31
Upgrade Electron ecosystem (37/EB 26) + deps, CI, React types

notluquis:
- Bumped Electron from 27.0.2 → 37.4.0
- Bumped electron-builder from 24.9.1 → 26.0.19
- Updated `isDevMode` implementation to use `!app.isPackaged` (safer, supported Electron API)
- Updated yarn.lock accordingly
- Keeps project aligned with the latest Electron ecosystem, including security patches and compatibility updates

Tested locally on:
• OS: macOS arm64 (Darwin 25, Apple Silicon)
• Launching JupyterLab and creating notebooks worked
• Kernel execution, inline plotting, ipywidgets, ipympl all confirmed


notluquis:
Also tried in Windows 11 x64, all works as expected.

ijstokes:
OK, I don't want to hold things up so why don't we agree on what we think are the minimal set of issues to address here and now?  From my side I'd say:

* Need environment prompting to work
* Need clean exit on quit
* Need to confirm code signing works OK

I don't think we need more than that, but please let me know your thoughts.  A "nice to have" would be to improve three points in the [dev.md](https://github.com/jupyterlab/jupyterlab-desktop/blob/master/dev.md) contributor description:

* how to setup a local "clean" dev & test environment (e.g. purge or at least hide any Python, Conda, Jupyter settings to maximally mimic a first-time-installation situation for a new user)
* how to smoke test a newly built instance
* a comment on finding CI builds in the [GH Actions](https://github.com/jupyterlab/jupyterlab-desktop/actions) CICD area.

Those are all documentation issues which could/should go in a separate PR, and are not specifically linked to this PR for Electron upgrade, however while you're putting some time into this it seems like you have your own techniques for 1 & 2 which would be good to capture for the benefit of others (e.g. me!).

Regarding your points:

* Code signing: I will check later today if the CI build for this PR [here](https://github.com/jupyterlab/jupyterlab-desktop/actions/runs/17368106367?pr=910) installs without any issues on macOS
* Bundled environment: I'll investigate more -- but if you have any ideas/pointers, please LMK
* Exit hang: not an issue if it doesn't happen with the CI build, and if it does I'll know where to look in the logs to investigate further.
* Electron 38: not a priority for this PR, but if I get everything else working I may see if there is a minimum delta on `packages.json` to bump it to 38 and see if that works.
* Other dependencies: push off until another PR.

PR #954, opened 2026-05-06
fix: replace node-fetch with Electron net.fetch in main process

### LINKED ISSUE #953: Welcome screen news feed not loading; TLS error fetching blog.jupyter.org
Since at least early 2026, the news feed in the welcome screen silently fails for users without a previously cached news list. Users with an existing `app-data.json` cache continue to see stale articles.

**Symptom in logs (`~/Library/Logs/jupyterlab-desktop/main.log`):**
```
Failed to fetch news list: FetchError: request to https://blog.jupyter.org/feed failed, reason: unable to verify the first certificate
```

**What I observed:**
`blog.jupyter.org` sends two certificates in the TLS handshake, but they don't form a complete chain — the leaf cert's issuer (`Sectigo Public Server Authentication CA DV R36`) is not the same as the intermediate that's being served (`Sectigo RSA Domain Validation Secure Server CA`). Node.js's TLS implementation fails to verify it. `curl` and browsers appear to handle it fine, possibly via AIA fetching or cached intermediates in the system store, I'm not certain of the exact mechanism.

Confirmed: cert issued Jan 27 2026, errors start from that date in logs.

`node-fetch` is used in two places in the main process (`welcomeview.ts` for the news feed, `app.ts` for update checks) — the update check may be affected too depending on GitHub's cert chain.

**To reproduce:**
1. Remove cached news list: set `newsList: []` in `~/Library/Application Support/JupyterLab/app-data.json`
2. Launch the app
3. Welcome screen shows only "Jupyter Blog" link — no articles
4. Log shows the TLS error above

**Possible fix:** replace `node-fetch` with Electron's `net.fetch` (available since Electron 21) which uses Chromium's network stack. Not sure if this is the right approach or if there's a preferred pattern here.

Note: AI-assisted (Claude Code).

notluquis:
Fixes #953.

`node-fetch` uses Node.js TLS which fails on incomplete certificate chains. `blog.jupyter.org` has served a broken chain since Jan 27 2026 — the leaf cert's issuer doesn't match the intermediate being sent. Node.js rejects it; Chromium handles it fine (AIA fetching / cached intermediates).

Replacing both `node-fetch` usages in the main process with `net.fetch` from Electron's `net` module, which uses the Chromium network stack. This also inherits system proxy settings.

**Changes:**
- `welcomeview.ts`: `net.fetch` for news feed; also adds `isArray` option to `XMLParser` for fast-xml-parser v5 compatibility (single `<item>` collapsed to object without it)
- `app.ts`: `net.fetch` for update check
- removes `node-fetch` and `@types/node-fetch` (no remaining usages)

**Testing:**
1. Cleared `newsList` in `app-data.json` to remove cache
2. Ran dev build with `node-fetch` — `FetchError: unable to verify the first certificate`, news section empty
3. Ran dev build with `net.fetch` — 10 articles loaded, no error
4. Build clean, 341 unit tests pass

Note: AI-assisted (Claude Code). Manually verified: reproduced bug and confirmed fix by running the dev build both ways and checking app-data.json cache state.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,14 +27,12 @@ jobs:
       - name: Build app
         run: yarn build
 
-      - name: Install Playwright browsers
+      - name: Install Playwright Chromium + system deps
         run: npx playwright install --with-deps chromium
 
-      # Electron needs a display on Linux
+      # xvfb-maybe auto-detects headless Linux and wraps with xvfb-run
       - name: Run E2E tests
-        uses: coactions/setup-xvfb@v1
-        with:
-          run: yarn test:e2e
+        run: yarn test:e2e
         env:
           CI: true
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,47 @@
+name: E2E Tests
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  e2e:
+    name: 'E2E smoke tests'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+          cache: 'yarn'
+
+      - name: Install dependencies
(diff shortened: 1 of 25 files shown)


notluquis:
Closing — PR was opened from wrong branch (included unrelated commits). Reopening from a clean branch.

Issue #951, opened 2026-05-05
Revival: Electron 41, test infra, deps, CI


notluquis:
This is a comprehensive tracking issue for reviving active maintenance of JupyterLab Desktop. Supersedes #919 (Electron v38, EOL).

**Work branch:** `phase3/electron-42` for Phase 3; Phase 1–2 PRs land directly on master. (The earlier `revival/electron-41` branch was retired: it carried obsolete Phase 2 commits already on master plus security work that belongs in the private advisory process, so Phase 3 starts from a clean branch off current master.)

**Current state:**
- Electron 27.0.2 — EOL, 15 majors behind (current: 42.x)
- JupyterLab at 4.5.7 (#950 merged)
- TypeScript 4.2 (current: 6.x); no React in the source, the `@types/react` devDeps are dead weight to drop
- Apple cert and CI secrets ~~expired during the maintenance gap~~ status unconfirmed from the code (no signed release since v4.2.5-1, 2024-08-29) — see To verify
- ~~macOS Tahoe perf regression on current Electron (electron/electron#48311)~~ moved to To verify (can't confirm applicability from the code)
- Several draft security advisories pending fix/publication before next release

Previous Electron upgrade attempt: PR #910 (Electron 37, self-closed) — not merged, not rebased onto current branch. Build issues documented in #919. Phase 3 starts from scratch.

---

## Phase 1 — Dependency security fixes

Triaged the Dependabot bumps by where each dep runs (`yarn why`). Build/dev-only deps don't ship in the asar (yarn 1 omits devDeps), so their runtime risk is null even when the PR is stuck.

- [x] `fast-xml-parser` 4.2.5 → 5.7.0 (XML entity fix) — #849 merged
- [x] `serialize-javascript` → 6.0.2 (XSS), build-only via `terser-webpack-plugin`; never reaches the build output on master
- [x] `webpack` 5.76 → 5.94 (#863, rebased, open/MERGEABLE) — direct devDep
- [x] Transitive bumps closed (#957 postcss, #956 xmldom, #923 tmp, #864 micromatch): fixed by bumping their parents (electron-builder, @typescript-eslint) in Phase 4, not via `resolutions`
- [ ] `js-yaml` 4.1.0 → 4.1.1 (#934, runtime path in `app.ts`) — rebased, awaiting review

> AI-assisted (Claude Code). Rebases verified locally with `yarn build`.

---

## Phase 2 — Test infrastructure

[...]

PR #960, opened 2026-05-08
ci: cross-platform test matrix and CI hardening

notluquis:
First of a four-part split for the testing/CI work tracked in #951. This one is CI infrastructure only; it does not add test files (those land in follow-up PRs B/C/D).

## Why split

The full plan is roughly 26 new tests + 1 packaging-smoke workflow split across four PRs (this one + e2e/unit/security/visual). Doing it as one mega-PR would be unreviewable. This PR is the foundation: cross-platform CI + safety guards. PRs B/C/D depend on having the matrix and bundle-size budget already in place.

## What this PR changes

| Area | Change |
|---|---|
| Unit tests | Run on Linux + macOS + Windows in parallel (was Linux only) |
| E2E smoke | Same 3-OS matrix (was Linux only); Playwright `--with-deps` only on Linux |
| Type check | Explicit `npx tsc --noEmit` step before build in both workflows |
| Bundle budget | New `scripts/check-bundle-size.js`, fails if any preload bundle exceeds 30 KB (current sizes 9.5 to 14.4 KB) |
| Concurrency | Cancel outdated PR runs per workflow; pushes to `master` queue normally so a release-tag push is never aborted |
| Permissions | Workflow-level `contents: read`, with `contents: write` override on the publish job that actually uploads release assets |
| Timeout | 15 min cap on the unit-test job (e2e already had 20 min) |
| Coverage upload | Linux only, to avoid duplicate artifacts from the matrix |
| Artifact naming | E2E test-results suffixed by `${{ matrix.os }}` so the three jobs don't collide on `actions/upload-artifact@v4` |
| Lint unblock | Pre-existing prettier drift on `test/` and `vitest.config.ts` was masking eslint output through the `prettier:check && eslint:check` chain. This PR formats those files (no source under `src/` is touched) so `yarn lint:check` exits 0 and the rest of the matrix can run |
| Docs | New "Automated test coverage" section in `dev.md` describing what runs |

## Commit-by-commit (squash on merge if you prefer)

1. `ci: run unit tests on Linux, macOS, and Windows`
2. `ci: run e2e smoke tests on Linux, macOS, and Windows`
3. `ci: add explicit tsc --noEmit type check step`
4. `ci: add preload bundle size budget check`
5. `docs(ci): document automated test coverage and matrix`
6. `ci: cancel outdated PR runs via concurrency group`
7. `ci: declare least-privilege permissions on workflows`
8. `ci: cap unit test job at 15 minutes`
9. `style: prettier-format test files and vitest config`

## Verification I ran

Local on macOS, against the merged tip of `revival/electron-41`:

- `python -c "import yaml; yaml.safe_load(...)"` on both workflows: parses cleanly.
- `npx tsc --noEmit`: exits 0.
- `yarn build`: webpack 5.76.0 compiles all 12 preloads.
- `node scripts/check-bundle-size.js`: 12/12 OK, all under 30 KB.
- `yarn test:unit`: 11 files, 341 / 341 pass.
- `yarn lint:check`: exits 0 after the prettier-format commit (22 pre-existing eslint warnings remain, all about unused imports and missing vitest globals; out of scope to fix here).

What I could not verify locally and is left to CI on this PR:

- Multi-platform behavior on Windows (I am on macOS).
- E2E suite (`yarn test:e2e`): Chromium needs the platform-specific deps that the workflow itself installs.
- Concurrency cancellation: only observable by force-pushing twice to this PR; the matrix run on this push is the first opportunity to confirm behavior.

## Caveats and judgement calls

- Bundle budget set at 30 KB per preload (about 2x the current largest, `pythonenvdialog` at 14.4 KB). Generous on purpose; intent is to catch regressions, not police every byte. Easy to lower later if you'd rather a tighter budget.
- I did not pin actions to commit SHA. The repo currently uses tags everywhere (`actions/checkout@v4`, etc.) and Dependabot covers `github-actions` weekly. SHA pinning is a separate security concern.
- The publish job in `publish.yml` got `contents: write` because it uploads release assets via `svenstaro/upload-release-action`. If you'd rather scope that even tighter (e.g. only when the release-exists step returns true), happy to follow up.
- Concurrency keys are `${{ github.workflow }}-${{ github.ref }}` so `publish.yml` and `e2e.yml` do not cancel each other.

## Out of scope (deferred to PRs B/C/D)

- New tests (e2e expansion, unit coverage gaps, security tests, visual regression).
- Packaging smoke workflow (`dist:<os> --publish never` per OS in PR).
- Coverage threshold enforcement (better fixed once the unit gap tests in PR-C give a real coverage number).

> Note: AI-assisted (Claude Code). Manually verified the seven local checks listed above and the YAML syntax. Each commit body has its own AI disclosure with the specific verification per change.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -34,6 +34,9 @@ jobs:
       - name: Build app
         run: yarn build
 
+      - name: Check preload bundle sizes
+        run: node scripts/check-bundle-size.js
+
       # --with-deps installs Linux apt packages; macOS/Windows handle deps natively
       - name: Install Playwright Chromium (Linux)
         if: runner.os == 'Linux'
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -11,6 +11,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+permissions:
+  contents: read
+
 jobs:
   e2e:
     name: 'E2E smoke (${{ matrix.os }})'
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -7,6 +7,10 @@ on:
     branches: [ master ]
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   e2e:
     name: 'E2E smoke (${{ matrix.os }})'
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,6 +28,9 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
+      - name: Type check
+        run: npx tsc --noEmit
+
       - name: Build app
         run: yarn build
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -9,8 +9,12 @@ on:
 
 jobs:
   e2e:
-    name: 'E2E smoke tests'
-    runs-on: ubuntu-latest
+    name: 'E2E smoke (${{ matrix.os }})'
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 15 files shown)


notluquis:
Closing — wrong target branch. Reopening against master per the phase-by-phase plan in #951.

PR #963, opened 2026-05-16
Disable yarn scripts

krassowski:
Just checking if we can make it a bit more secure

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+enableScripts: false


notluquis:
Isn't this repo still on Yarn 1? The lockfile header is "yarn lockfile v1" and CI installs yarn with npm install --global yarn, so I think .yarnrc.yml (Yarn 2+ config) just gets ignored here. I tried adding it locally and yarn install still ran the build scripts.

For Yarn 1 the closest equivalent seems to be yarn install --ignore-scripts, but that blocks all scripts including ours, not only dependencies. Were you thinking of a Berry migration for this?

krassowski:
Oh indeed, that's ancient:

https://github.com/jupyterlab/jupyterlab-desktop/blob/9549bc1531267236f3ea0b7d9d1117e41188e841/yarn.lock#L2

Good call. Let's think about other ways to make it more secure/resilient and upgrade yarn to berry at some later time.

notluquis:
@krassowski sounds good. A few resilience wins that work on Yarn 1 today, leaving the Berry migration for later:

- Build provenance, since we ship signed installers, adding actions/attest-build-provenance to the release job would let users verify a .dmg/.exe/.deb actually came from this CI. That is probably the highest-value supply-chain step here once releases resume.
- Pin the workflow actions to commit SHAs instead of tags, starting with the third-party ones in publish.yml (it carries the signing/release secrets, so a moved tag there is the higher risk). Dependabot's github-actions config already keeps SHA pins updated, so it stays low-maintenance.
- Set a contents: read default token at the workflow level; the test job has it but the publish job currently inherits the repo default.
- Use --frozen-lockfile on the build and sync install steps too, not just the unit job.

Smaller extra: the only install script in the current tree is electron's own postinstall, so yarn install --ignore-scripts on the unit test job (electron is mocked there) blocks install-time execution on that runner cheaply, without touching the build/publish jobs that need it.

I can open a small hardening PR with whichever of these you like, and revisit Berry plus a per-dependency script allowlist when the build stack moves in a future phase.

PR #968, opened 2026-05-26
build: husky + lint-staged pre-commit hook (C2b)

notluquis:
Part of #951 Phase 2 (C2b). Adds a pre-commit hook so formatting and the relevant tests run before a commit lands, not only in CI.

## What it does
- `prepare: husky` installs the hook on `yarn install`.
- `.husky/pre-commit` runs `lint-staged`.
- lint-staged config: `*.{ts,tsx,js,jsx}` -> `eslint --fix`, `prettier --write`, `vitest related --run`; `*.{json,md,css}` -> `prettier --write`.

`vitest related --run` runs only the tests that import the staged files, so the hook stays fast on a small change instead of running the whole suite.

husky and lint-staged were already in `node_modules` but missing from `package.json` and the lockfile; this adds both (and the lockfile entries, so `yarn install --frozen-lockfile` stays green).

## Verified
`yarn install` and `yarn install --frozen-lockfile` both exit 0 and write the deps to `yarn.lock`. Staged a deliberately mis-formatted source file and confirmed eslint, prettier and `vitest related` all run; the commit of this PR itself went through the hook.

Note: AI-assisted (Claude Code). Manually verified: the install and hook behavior described above, on this branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "./build/out/main/main.js",
   "scripts": {
     "start": "electron .",
+    "prepare": "husky",
     "test": "yarn test:unit",
     "test:unit": "vitest run",
     "test:unit:watch": "vitest",
@@ -55,6 +56,16 @@
     "prettier": "prettier --write \"**/*{.ts,.tsx,.js,.jsx,.css,.json,.md}\"",
     "prettier:check": "prettier --check \"**/*{.ts,.tsx,.js,.jsx,.css,.json,.md}\"",
     "update-ufuzzy": "shx cp node_modules/@leeoniya/ufuzzy/dist/uFuzzy.iife.min.js src/assets/"
+  },
+  "lint-staged": {
+    "*.{ts,tsx,js,jsx}": [
+      "eslint --fix",
+      "prettier --write",
+      "vitest related --run"
+    ],
+    "*.{json,md,css}": [
+      "prettier --write"
+    ]
   },
   "build": {
     "appId": "org.jupyter.jupyterlab-desktop",
@@ -205,7 +216,9 @@
     "eslint-plugin-prettier": "~4.0.0",
     "eslint-plugin-react": "~7.29.4",
     "fs-extra": "~9.1.0",
+    "husky": "^9.1.7",
     "istextorbinary": "^6.0.0",
+    "lint-staged": "^16.4.0",
     "meow": "^6.0.1",
     "node-watch": "^0.7.4",
     "prettier": "~2.1.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -205,7 +216,9 @@
     "eslint-plugin-prettier": "~4.0.0",
     "eslint-plugin-react": "~7.29.4",
     "fs-extra": "~9.1.0",
+    "husky": "^9.1.7",
     "istextorbinary": "^6.0.0",
+    "lint-staged": "^16.4.0",
```
krassowski, line 221: Do we need to add more `npm` dependencies?

I am somewhat -1 on the proposed change; if anything I would much prefer using `pre-commit` from Python ecosystem instead because this is what most of Jupyter repos are using and it does not pull extra 203 lines into the lockfile.

notluquis: Fair point, the lockfile growth is the part I like least about it too.

I had a look at how prettier and eslint get run elsewhere in the org before reworking anything, and it's pretty mixed. jupyterlab runs them as pre-commit `local` `language: node` hooks (with `default_language_version: node: system` and the hooks under `ci.skip`); jupyter-collaboration mirrors that, and voila uses the same local-node approach. On the other hand lumino, ipywidgets, and the TS side of jupyterlab-git use husky plus lint-staged, with jupyterlab-git keeping pre-commit for black only. So both patterns have precedent, but jupyterlab's feels like the closest thing to a house style to follow here.

If you're happy with that, I'd switch this PR to a pre-commit config mirroring jupyterlab's instead of husky/lint-staged. It adds nothing to yarn.lock, and I'd keep tests out of the hook (CI still runs the full vitest suite), so it would only do formatting and lint. Something like:

```yaml
ci:
  skip: [prettier, eslint]
default_language_version:
  node: system
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v5.0.0
    hooks:
      - id: trailing-whitespace
      - id: end-of-file-fixer
  - repo: local
    hooks:
      - id: prettier
        name: prettier
        entry: prettier --write
        language: node
        types_or: [json, ts, tsx, javascript, jsx, css, markdown]
      - id: eslint
        name: eslint
        entry: eslint --fix
        language: node
        types_or: [ts, tsx, javascript, jsx]
```

pre-commit passes the staged filenames to each entry, so these call the binaries directly rather than the repo-wide `yarn prettier` / `yarn eslint` scripts (same idea as jupyterlab's `prettier:files` / `eslint:files`, where `eslint:files` is literally `eslint --fix`).

Would that work for you, or would you rather this repo not carry a commit hook at all and leave it to CI? Either is fine by me, just let me know which way you'd prefer.

PR #970, opened 2026-05-27
test: characterize the 12 preload contextBridge contracts

notluquis:
Part of the #951 revival work (Phase 2 test coverage), ahead of the Electron upgrade in Phase 3.

The 12 `src/main/*/preload.ts` scripts had no unit coverage. Since Phase 3 rewrites them under the Electron 29+ `contextBridge` rules, I wanted the current contract pinned first so the rewrite has a regression net rather than a manual recheck.

## What it does
Adds one characterization test per preload. Each pins:
- the exact set of keys exposed under the `electronAPI` namespace,
- the `EventTypeMain` channel and arguments each function forwards through `ipcRenderer.invoke`/`send`,
- the `EventTypeRenderer` channels each listener registers and relays to its callback,
- a guard that the exposed API does not contain the raw `ipcRenderer` object (the Electron 29 change where passing it whole yields an empty object). All 12 already wrap calls, so this passes today and should stay green after the migration.

## Testing change
The preloads use `require('electron')`. vitest leaves `electron` externalized, so that require returns the binary path string and bypasses `vi.mock`, which is why preload code wasn't unit-testable. I moved the existing inline `electron` mock into a single `test/setup/electron-stub.ts` reached two ways that share the same `vi.fn` instances: a `resolve.alias` for ESM imports, and a `require.cache` seed for the CJS require. The existing suite is unaffected.

## Verified
Full unit suite green (409 tests, the prior 320 plus 89 here). `tsc --noEmit`, eslint and prettier clean. Mutation check: flipping a forwarder channel (aboutdialog) and a listener channel (titlebarview) makes the relevant tests fail as expected, so they are not tautological.

Note: AI-assisted (Claude Code). Manually verified: the runs described above (full suite, tsc, lint, and the two mutation checks) on this branch.

### DIFF
--- a/test/setup/electron-mock.ts
+++ b/test/setup/electron-mock.ts
@@ -1,48 +1,27 @@
 import { vi } from 'vitest';
 import { tmpdir } from 'os';
 import { join } from 'path';
+import * as electronStub from './electron-stub';
 
-const userDataPath = join(tmpdir(), 'jlab-test-userdata');
 const logFilePath = join(tmpdir(), 'jlab-test.log');
 
-vi.mock('electron', () => ({
-  app: {
-    getPath: vi.fn((name: string) => join(userDataPath, name)),
-    getVersion: vi.fn(() => '4.4.7'),
-    getName: vi.fn(() => 'JupyterLab'),
-    isPackaged: false,
-    whenReady: vi.fn(() => Promise.resolve())
-  },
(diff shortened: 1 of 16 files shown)


notluquis:
Reopening from a branch on the repo itself to match the Phase 2 convention, instead of from a fork. Same commits.

PR #972, opened 2026-05-27
test: cover cli.ts and env.ts process-spawn branches

notluquis:
Part of the #951 revival work (Phase 2 test coverage), ahead of the Electron/Node upgrade in Phase 3.

C2 (#967) deliberately left the `spawn`/`execFileSync` and Python-discovery branches of `env.ts` and `cli.ts` uncovered (noted in the vitest config comment). Those are exactly the paths the Phase 3 Node ABI jump (Node 18 to 24) and Electron upgrade touch, so a regression there would surface at runtime rather than as a test failure. This adds unit coverage for them.

## env.ts (`test/unit/env-spawn.test.ts`, 16 tests)
- `runCommandInEnvironment`: spawns `bash -c` (and `cmd /c` on Windows), resolves true/false on the close code, relays stdout to the callback.
- `validatePythonPath` / `validateSystemPythonPath`: execFileSync success and the not-exist / wrong-output / throw paths.
- `validateCondaPath`: conda info JSON parsing, the not-base-env branch, and missing `conda_version`.
- `getEnvironmentInfoFromPythonPath` and its sync variant: env-type mapping and the failure path.

## cli.ts (`test/unit/cli-spawn.test.ts`, 5 tests)
- `launchCLIinEnvironment`: spawns bash with the generated activate script and unlinks the temp file on close.
- `createPythonEnvironment`: conda-lock-file source runs conda-lock in the base conda env; the venv source uses execFileSync on the system python; the missing-base-conda and missing-python branches throw.

Mocks stay at the boundaries (`child_process`, `fs`, and the `env`/`utils` modules), mirroring the existing `env.test.ts` and `cli-handlers.test.ts` style.

## Verified
Full unit suite green (341 tests) with `yarn test:coverage`, so the existing CI-enforced thresholds still pass; env.ts branch coverage rises to ~43%. `tsc --noEmit` clean. Mutation checks: flipping the `bash -c` argument and the venv `execFileSync` args each fail the relevant test, so they are not tautological.

Note: AI-assisted (Claude Code). Manually verified: the runs above (coverage suite, tsc, and the two mutation checks) on this branch.

### DIFF
--- a/test/unit/cli-spawn.test.ts
+++ b/test/unit/cli-spawn.test.ts
@@ -1,0 +1,156 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import * as fs from 'fs';
+import { execFileSync, spawn } from 'child_process';
+
+vi.mock('fs', async () => {
+  const actual = await vi.importActual<typeof import('fs')>('fs');
+  return {
+    ...actual,
+    existsSync: vi.fn(() => true),
+    unlinkSync: vi.fn()
+  };
+});
+vi.mock('child_process', async () => {
+  const actual = await vi.importActual<typeof import('child_process')>(
+    'child_process'
+  );
+  return { ...actual, execFileSync: vi.fn(), spawn: vi.fn() };
+});
+vi.mock('../../src/main/env', () => ({
+  getCondaPath: vi.fn(() => '/conda/condabin/conda'),
+  getSystemPythonPath: vi.fn(() => '/usr/bin/python3'),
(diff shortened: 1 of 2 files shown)


notluquis:
Ready when you have time, @krassowski. Extends `cli.ts` and `env.ts` coverage into the spawn paths C2 left uncovered (Python discovery, validators, command execution), boundary mocks only on `child_process` and `fs`.

krassowski (approved):
Thanks!

PR #974, opened 2026-05-27
ci: add a preload bundle size budget check

notluquis:
Part of the #951 revival work (Phase 2), groundwork for the Phase 3 Electron upgrade.

The 12 preload scripts are bundled by webpack. Phase 3 rewrites all of them under the Electron 29+ contextBridge rules, so I wanted a cheap tripwire for an unexpected size jump (or a bundle silently being added or removed).

## What it does
`scripts/check-preload-budget.js` (no dependencies) reads each `build/out/main/*/preload.js`, compares it to a per-file byte ceiling set ~20% over the current built size, and exits non-zero if any bundle is over budget, missing, or present but not in the budget map. `bundle-budget.yml` runs `yarn build` then the script on push/PR to master.

I went with a small node script rather than `size-limit` on purpose: size-limit's gzip/transfer modeling is aimed at web bundles and would add devDependencies, whereas here a raw file-size ceiling is all that's needed.

## Verified
The check passes at current sizes, and fails when a ceiling is lowered below the built size. Actions are SHA-pinned with a `# vX` comment matching publish.yml. New workflow, so it won't gate merges until added as a required check.

Note: AI-assisted (Claude Code). Manually verified: the pass and the deliberate-regression failure described above, locally.

### DIFF
--- a/.github/workflows/bundle-budget.yml
+++ b/.github/workflows/bundle-budget.yml
@@ -1,0 +1,28 @@
+name: Preload bundle budget
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  bundle-budget:
+    name: 'preload size budget'
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


notluquis:
Dropping this. For a desktop app preload bundle size has no transfer cost, and actual compile/bundle breakage is already caught by the installer build, so the size budget adds little. Focusing the C3 test work on the type-check, e2e and characterization detectors instead.

PR #978, opened 2026-06-14
Phase 3 pre-bump: TypeScript 6 and ESLint 10 flat config

notluquis:
Part of #951 Phase 3. The type stack has to move before the Electron 42 bump can land: Electron 42 runs on Node 24, which forces @types/node up, which forces TypeScript up (@types/node@24 declares typeScriptVersion 5.3, so TS 4.2 cannot compile it; typescript-eslint 8 supports TS up to <6.1). So this rides ahead of the bump.

What it does:
- TypeScript 4.2 to 6.0, @types/node 14 to 24
- ESLint 8 to 10 with the flat-config migration (.eslintrc.js to eslint.config.mjs, tseslint.config())
- removes the dead @types/react, @types/react-dom, eslint-plugin-react (no React in src)
- adds a fast tsc --noEmit gate workflow (publish.yml only runs tsc inside the slow installer job)
- the 12 preloads move to import electron = require('electron') for native ipcRenderer types; runtime stays a require (sandbox-compatible)
- fast-xml-parser 5.7 to 5.8, which fixes an upstream .d.cts typo that blocked nodenext resolution

moduleResolution is nodenext, strict stays off to preserve the 4.2 level (the strict-null migration is a separate later step), skipLibCheck is on: electron 27 ships its own electron.d.ts whose `noDeprecation` declaration conflicts with @types/node 24 (TS2687, "All declarations of 'noDeprecation' must have identical modifiers"), so the lib graph cannot be fully checked until the Electron bump moves electron itself.

One thing I want to flag rather than bury: the old setup had two non-CI advisory scripts (eslint:check:typed / eslint:typed) that ran .eslintrc.typecheck.js to enable type-aware rules, mainly no-floating-promises. They were never wired into lint:check or any workflow. I did not carry them into the flat config. Measuring under a typed flat-config block, no-floating-promises currently reports 84 hits across src/main (this matches the pre-existing state, the migration does not add any), so restoring it as an error would be its own cleanup PR, and restoring it as an always-red advisory script felt like dead weight. I would rather defer type-aware linting to a dedicated follow-up that also fixes the 84 sites, but happy to instead re-add the separate advisory script here if you prefer parity.

Note: AI-assisted (Claude Code). Manually verified: tsc 0 errors, eslint and prettier clean, full build, the unit suite, and the e2e suite all green locally; measured the no-floating-promises count (84) by adding a temporary projectService typed block and running eslint 10.5.0 / typescript-eslint 8.61.0 on src/main, then reverted it.


notluquis:
ready for review @krassowski

krassowski (approved):
Thanks!

PR #979, opened 2026-06-14
Phase 3 e2e: macOS leg and embedded-server isolation

notluquis:
Part of #951 Phase 3 e2e hardening. Keeps the e2e suite scoped to the desktop shell plus a single integration test, rather than re-testing JupyterLab itself (that is upstream's job via Galata).

What the suite covers after this PR:
- smoke: the app launches, the welcome window renders, the views compose, and it shuts down cleanly. No server, no Python needed.
- python-env (from #976): clicking New notebook boots a real Jupyter server and the labview reaches the server's /lab URL. The one integration test, proving the shell correctly wraps a server.

This PR adds:
- A macOS leg to the e2e matrix alongside Linux (the shipped platforms). On the macOS runner extract-zip leaves an incomplete Electron.app, so the leg re-extracts the cached zip with ditto (resolved via @electron/get) before launch.
- Per-launch isolation of the embedded server: HOME and the user-layer Jupyter dirs are redirected to temp dirs so a test that boots the server never reads or writes the runner's real ~/.jupyter. CONFIG is set via JLAB_DESKTOP_CONFIG_DIR (the app's own override, which server.ts honors) rather than a JUPYTER_CONFIG_DIR the app would ignore.
- A shared pageByUrl wait (polls every open page's URL, catching an in-place navigation as well as a new window) and a shared LAB_URL / NEEDS_PYTHON.

Note: AI-assisted (Claude Code). Manually verified: built and ran the suite locally against jupyterlab 4.5.6 with --workers=1 (the CI setting), 5 passed; eslint clean; the app opens no external browser (it renders JupyterLab in the WebContentsView).


notluquis:
Superseded by #984, which has the same change with a clean two-commit history (this branch accumulated the macOS CI repair debugging and the test add/remove churn). Closing in favor of #984.

PR #980, opened 2026-06-14
Phase 3 pre-bump: explicit hardened webPreferences at every window

notluquis:
Part of #951 Phase 3 pre-bump PRs. De-risks the Electron 42 bump and is safe at Electron 27.

Explicit webPreferences (contextIsolation, nodeIntegration, sandbox) at every BrowserView/BrowserWindow construction site via a shared hardenedWebPreferences helper, instead of relying on Electron defaults. Closes the nodeIntegration/contextIsolation advisory and prevents a silent regression when Electron 42 changes the defaults. sandbox:true is verified non-breaking: the 12 preloads only use contextBridge/ipcRenderer.

This PR originally also carried a JLAB_DESKTOP_E2E_MODE opt-out and a stub-env-tar script. Both were dropped: the single-instance lock is keyed on the userData dir so a unique --user-data-dir per e2e already isolates it (no opt-out needed), the relaunch sites are user-triggered, and neither piece had a consumer yet. They will return with the packaged-build e2e in a later phase, where a real launch proves them necessary. The CodeQL file-system-race finding went away with the stub-tar test that contained it.

Note: AI-assisted (Claude Code). Manually verified: tsc --noEmit clean, unit suite green, coverage over the floor, eslint clean; confirmed the single-instance handoff works natively by launching two instances against one userData dir (the second logs "Handling request in the main instance." and exits).


notluquis:
Closing this. After dropping the e2e-mode opt-out and the stub-env-tar from this PR, the only remaining change was adding explicit contextIsolation/nodeIntegration/sandbox at the window construction sites via a hardenedWebPreferences helper.

Checked the defaults against the Electron security tutorial: nodeIntegration is false by default since 5.0, contextIsolation true since 12.0, and sandbox true since 20.0. At Electron 27 (and 42) all three secure values already apply, and the security checklist is satisfied by the defaults rather than requiring the flags to be written out. master already relies on the defaults at every one of these sites and is green, so the explicit declaration here does not change runtime behavior.

Rather than land a helper that only restates the defaults (and that a static analyzer like Electronegativity cannot resolve through a function call), I would rather rely on the defaults like master does. If defense-in-depth explicit flags are wanted later on the remote-content surfaces (labview, the connect window), that is a small focused change we can do on its own with that rationale.

Note: AI-assisted (Claude Code). Manually verified the default values against the Electron docs and that there are no open CodeQL alerts for these sites.

PR #984, opened 2026-06-16
Phase 3 e2e: macOS leg and embedded-server isolation

notluquis:
Part of #951 Phase 3 e2e hardening. Replaces #979 with a clean two-commit history (same change, no churn).

Keeps the e2e suite scoped to the desktop shell plus a single integration test, rather than re-testing JupyterLab itself (that is upstream's job via Galata).

The suite after this PR:
- smoke: the app launches, the welcome window renders, the views compose, and it shuts down cleanly. No server, no Python needed.
- python-env (from #976): clicking New notebook boots a real Jupyter server and the labview reaches the server's /lab URL. The one integration test, proving the shell correctly wraps a server.

This PR adds:
- A macOS leg to the e2e matrix alongside Linux. On the macOS runner extract-zip leaves an incomplete Electron.app, so the leg re-extracts the cached zip with ditto (resolved via @electron/get) before launch.
- Per-launch isolation of the embedded server: HOME and the user-layer Jupyter dirs are redirected to temp dirs so a test never reads or writes the runner's real ~/.jupyter. CONFIG is set via JLAB_DESKTOP_CONFIG_DIR (the app's own override) rather than a JUPYTER_CONFIG_DIR the app would ignore. Adds a unit test on server.ts for that override.
- A shared pageByUrl wait (catches an in-place navigation as well as a new window) and a shared LAB_URL / NEEDS_PYTHON.

Note: AI-assisted (Claude Code). Manually verified: built and ran the suite locally against jupyterlab 4.5.6 with --workers=1, 5 passed; both Playwright legs green on CI with 0 flaky; the app opens no external browser.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -15,8 +15,12 @@ concurrency:
 
 jobs:
   e2e:
-    name: 'Playwright (Linux)'
-    runs-on: ubuntu-latest
+    name: 'Playwright (${{ matrix.os }})'
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     # Bound the job so a stuck Electron launch/close fails fast instead of
     # tying up a runner; this suite specifically guards shutdown hangs.
     timeout-minutes: 15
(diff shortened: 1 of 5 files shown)


krassowski (approved):
Thanks

PR #991, opened 2026-06-28
build: refresh the vendored uFuzzy to match @leeoniya/ufuzzy 1.0.19

notluquis:
**What changes:** the contents of `src/assets/uFuzzy.iife.min.js`, from the old build to the 1.0.19 dist.

**Why:** the welcome view's environment search loads this vendored IIFE from `src/assets` directly, it does not import the npm package. #988 bumped `@leeoniya/ufuzzy` to 1.0.19 in `package.json` but left the vendored file at its previous build, so what ships and what the lockfile claims drifted apart. This is the kind of mismatch that only shows up when the bundled minified behavior differs from the pinned version.

**How:** regenerated through the existing `update-ufuzzy` script, which copies `node_modules/@leeoniya/ufuzzy/dist/uFuzzy.iife.min.js` into `src/assets`.

Written with Claude Code. I checked that master's dependency is on 1.0.19 while the committed vendored file was the older build, then launched the app to confirm the welcome search box still filters.


notluquis:
Closing in favor of #995. This PR just re-synced the vendored `src/assets/uFuzzy.iife.min.js` by hand, which is the same step that drifted in the first place (the dep was bumped to 1.0.19 but the committed copy stayed behind). #995 removes the vendored file and the `update-ufuzzy` script and copies uFuzzy from `node_modules` at build time instead, the same way `toolkit.min.js` is already handled, so the pinned dependency is the single source of truth and it cannot fall out of sync again.

PR #996, opened 2026-06-28
ci: open the release PR with gh instead of peter-evans/create-pull-request

notluquis:
#867 was a Dependabot bump of `peter-evans/create-pull-request`. @krassowski's note there was that we should drop the third-party action and use `gh pr` instead, so this does that rather than bumping it.

**What changes:** the last two steps of `releasepr.yml` (the throwaway change plus the `peter-evans/create-pull-request` call) become one `run` step that uses `git` and `gh pr create`. The checkout gains `token: JLAB_APP_TOKEN`, <s>and the job gets `contents: write` and `pull-requests: write`.</s>

**Why this shape:** `sync_lab_release.yml` already opens its PR with `git` + `gh pr create`, so this is not a new pattern, it is the one already in the repo. Removing peter-evans takes one more third-party action out of the release path.

**What is kept identical**, so the release flow does not move: the `workflow_dispatch` trigger, the pre-release-exists guard, the `release-v<version>` branch name, the `Update auto-release logs` commit, and the `Release v<version>` title and body. It still authenticates with `JLAB_APP_TOKEN`, which matters because a PR opened with the default `GITHUB_TOKEN` would not trigger CI.

**One real difference worth calling out:** peter-evans updated an existing PR in place on a re-run. Here `git checkout -b` fails if the branch already exists, so a re-run stops instead of touching an open PR. That is the same fail-safe `sync_lab_release.yml` relies on, so it is consistent with how the repo already does this, but it is a behavior change from peter-evans, not a like-for-like.

This closes #867, since the action it bumps is no longer referenced.

Written with Claude Code. I confirmed the YAML parses and that `peter-evans` is gone from `.github`, and the step is modeled on the working `gh pr create` block in `sync_lab_release.yml`. The workflow only runs on manual dispatch against an existing pre-release with the app token, so it cannot be exercised from a fork PR. That part I could not run, only read against the existing one.


.github/workflows/releasepr.yml
```diff
@@ -9,10 +9,15 @@ permissions:
...
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
     steps:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
-      
+        with:
+          token: ${{ secrets.JLAB_APP_TOKEN }}
```
krassowski: I think this can be done more securly, we can checkout with read-only permissions and only just before push use the token with write access

krassowski: Also `persist-credentials: false`. Do we have zizmor setup on this repo already? Maybe let's start with that, you can copy-paste https://github.com/jupyterlab/jupyterlab/blob/main/.github/workflows/zizmor.yml

.github/workflows/releasepr.yml
```diff
@@ -34,18 +39,32 @@ jobs:
...
-          title: 'Release v${{ steps.package-info.outputs.version}}'
-          body: |
-            Release v${{ steps.package-info.outputs.version}}
-          draft: false
+          date +%s > auto-release.log
+          git add auto-release.log
+          git commit -m "Update auto-release logs"
+          # Checkout persisted no credentials, so authenticate this one push with
+          # the app token instead of keeping it in the git config all job long.
+          git push \
```
krassowski: Could possible use `gh auth setup-git` for gh token setup. But I think this works too

notluquis:
Switched to `gh auth setup-git`, agreed it's cleaner. Checked that `gh auth git-credential` returns the token from GH_TOKEN in the runner so the push authenticates fine. Kept persist-credentials: false and moved the token to the push step.

.github/workflows/releasepr.yml
notluquis: Done both: the checkout now runs with persist-credentials: false and no job-level write, and the app token is only wired in right before the push via gh auth setup-git. On zizmor, there was nothing set up yet, so I ported the jupyterlab workflow you linked as a separate PR (#1016, now merged) running in reporter mode. This release workflow comes back clean under it, the only artipacked finding left was the checkout here, which persist-credentials: false clears.

krassowski (approved):
Let's try this

PR #997, opened 2026-06-28
ci: update the pinned actions to their current releases

notluquis:
Updates the SHA-pinned actions in the workflows to their current releases: `checkout` v7, `setup-node` v6, `github-script` v9, `upload-artifact` v7, `download-artifact` v8. The old pins still run on Node 20, so every job has been printing the runner's Node 20 deprecation warning; this clears that too.

These go further than the open Dependabot bumps (#917, #926, #929, #932), which were opened a while ago and propose older targets (v8, v5, v6, v6). Rather than land those and bump again later, this takes each one to the latest in a single change. Dependabot closes its own PRs once the same dependency lands on the default branch, so #917, #926, #929 and #932 should close themselves when this merges. This also covers `upload-artifact`, which has no Dependabot PR but was on the same Node 20 pin.

I read each changelog against how the workflows actually use the action, and none of the breaking changes apply here:

- `checkout` v7 blocks checking out a fork PR in `pull_request_target` and `workflow_run` workflows. We don't have either trigger.
- `github-script` v9 is ESM-only and drops `require('@actions/github')`. The three scripts use the injected `github` and `context` objects, no `require` and no `getOctokit`.
- `setup-node` v6 narrows automatic caching to npm. We set `cache: yarn` explicitly and there is no `packageManager` field in package.json, so nothing changes.
- `upload-artifact` v7's direct (unzipped) upload is opt-in via `archive: false`; every step here uploads by name and path.
- `download-artifact` v8 stops unzipping non-zip files and turns a digest mismatch into an error. Both downloads are name-based and the artifacts are the default zipped uploads.

One thing worth a look: `checkout` v6 changed where it stores the persisted git credentials. The only place that pushes after checkout is the `sync_lab_release` PR job, which passes an explicit `token:` and isn't a container action, so the push still works on the hosted runners. Calling it out in case I've missed a case.

Note: AI-assisted (Claude Code). I read each action changelog against how the workflows use it, confirmed the YAML parses and the pins resolve to the latest tags; the bump itself is mechanical.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -32,8 +32,8 @@ jobs:
       # on CI runners, so skip it and install the binary deterministically below.
       ELECTRON_SKIP_BINARY_DOWNLOAD: '1'
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
-      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
           # Match the Node that the target Electron bundles (Electron 42 -> Node
           # 24); do not run ahead of it. Node 26 broke Electron's install tooling.
@@ -92,7 +92,7 @@ jobs:
           else
             yarn test:e2e
           fi
-      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         if: ${{ !cancelled() }}
         with:
           name: playwright-report-${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


notluquis:
Closing this in favor of letting Dependabot handle it. There is a dependabot.yml that already updates github-actions weekly, so a manual bump PR duplicates that. The stale Dependabot action PRs were just lagging behind; `@dependabot recreate` brings each one up to the current latest (#926 is now checkout 7.0.0, #932 is download-artifact 8.0.1, #929 setup-node 6 and #917 github-script 9 are recomputing), so they cover the same ground and keep Dependabot as the single source for action versions. upload-artifact has no open PR yet but Dependabot manages it too (it did 3->4 in #741) and will open one on its next weekly run.

For whoever reviews those PRs, the changelog check I did while preparing this, none of the breaking changes affect us:
- checkout v7 only blocks fork-PR checkout in pull_request_target / workflow_run, which we don't use; the sync_lab_release push uses an explicit token so v6's credential-storage change is fine.
- github-script v9 is ESM-only and drops require('@actions/github'); the scripts use the injected github/context, so no change.
- setup-node v6 narrows auto-cache to npm; we set cache: yarn explicitly and have no packageManager field.
- upload v7 / download v8 only change opt-in direct uploads and non-zip handling; every step uses plain name/path.

Note: AI-assisted (Claude Code). Read each changelog against the workflows' usage; closing in favor of the Dependabot PRs.

PR #998, opened 2026-06-29
ci: have Dependabot track npm version updates too

notluquis:
The Dependabot config only listed `github-actions`, so on the npm side only the automatic security PRs ever appeared, routine version bumps were never proposed. This adds the `npm` ecosystem (which covers yarn too, Dependabot reads the `yarn.lock`) on a weekly schedule, with minor and patch bumps grouped into one PR so the queue doesn't flood.

Only one major is ignored: `@types/node`, because it should track the Node that Electron ships (24 on Electron 42) rather than the newest types. I deliberately left `yargs` in even though 18 is ESM-only and we can't merge it yet, keeping that bump visible is the reminder for when we do the ESM migration.

AI-assisted with Claude Code. The ignore rationale is mine; I only relied on tooling to confirm the yaml parses.


krassowski (commented):
Can you add 7-days cooldown?

notluquis:
@krassowski btw I've lost push access to the repo, my perms show pull-only now and git returns 403, though I was pushing PRs fine two days ago. Could you re-add me as a collaborator? Assuming it just lapsed. Can't update any PRs until then.

krassowski:
Sorry about that, yeswterday we were changing some security settings and I think GitHub did something weird. Anyways I sent an invite to add you back; regardless I think it is better to send PRs from forks in general.

krassowski (approved):

PR #999, opened 2026-06-29
build: drop @types/tar, tar 7 ships its own types

notluquis:
`tar` 7 has shipped its own type declarations since the Electron 42 bump landed, so the separate `@types/tar` devDependency is just dead weight now. Removing it, the types resolve straight from `node_modules/tar/dist` and tsc doesn't notice.

You can check this one quickly: drop the dep, `yarn install`, and `yarn type-check` stays at zero errors (the only tar usage is the `tar.x` call in the env extract). That's the run I did before opening this; AI-assisted with Claude Code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,7 +192,6 @@
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^24",
     "@types/semver": "^7.7.1",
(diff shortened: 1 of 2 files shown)


krassowski (approved):
Nice, thanks!

PR #1017, opened 2026-07-02
Harden the workflows flagged by zizmor

notluquis:
## References

- Findings come from the zizmor reporter added in #1016, which will show them in the Security tab.
- `releasepr.yml` is hardened in #996, so it is not touched here. The push pattern below is the same one #996 uses.

## Code changes

- **App token scoping** in `sync_lab_release`: the two `create-github-app-token` steps were inheriting every installation permission. Scoped them to what each job uses, read for the version check, `contents` + `pull-requests` write for the job that opens the update PR.
- **`persist-credentials: false`** on the checkouts in `e2e`, `typecheck`, `prerelease`, `publish` and `sync_lab_release`, so the default token is not left in the git config. The publish/prerelease release steps use `repo_token` through their actions rather than git push, so this is safe there. The one job that still pushes (the sync update PR) gets the app token via `gh auth setup-git` just for the push.
- **Template injection**: moved the JupyterLab version from a `${{ }}` expansion inside a run block into an `env:` var.

The only findings left after this are 10 informational `superfluous-actions` notes, which do not touch security. Those can follow up before flipping zizmor from report-only to a hard gate.

## User-facing changes

None. Workflow files only, nothing in the shipped application.

None.

## Manual testing

Nothing in `dev.md`'s review guidance applies, since no application code changes. Verified with zizmor locally that the app-token, artipacked and template-injection findings are gone.

## AI usage


notluquis:
Merged master in to clear the conflicts after the release-workflow changes (#1066, #1067, #1069).

Two things shrank in the process: master already picked up the `permission-contents: write` scoping on the prerelease app token, and `releasepr.yml` and `zizmor.yml` were added with `persist-credentials: false` already set. What is left here is the same set as before minus those: `persist-credentials: false` on the checkouts in `e2e`, `typecheck`, `prerelease`, `publish` and `sync_lab_release`, the app-token scoping in `sync_lab_release`, and the template-injection fix moving `LATEST` into `env:`.

Merged rather than rebased so the earlier review context stays intact.

PR #1020, opened 2026-07-02
ci: build the osx-64 installer on a native Intel runner

### LINKED ISSUE #1018: CI: macOS x64 installer flaky under Rosetta on arm64 macos-latest (conda pack Abort trap 6)
The `macOS x64 installer` job in the Publish workflow fails intermittently at the `Create Application Server Installer` step. When it fails, `conda pack` aborts like this:

```
$ conda pack -p ./env_installer/jlab_server -o ./env_installer/jlab_server.tar.gz
rosetta error: unable to mmap __TEXT: 84
 .../_csv.cpython-311-darwin.so.aot/bin/sh: line 1: 25563 Abort trap: 6   conda pack ...
error Command failed with exit code 134.
```

The install and the extras copy just before it succeed, so it is only the pack step that dies, with a `rosetta error` and `Abort trap: 6` (exit 134).

I think the cause is the runner image move rather than anything in our scripts. `macos-latest` is migrating to macOS 26 on Apple Silicon (actions/runner-images#14167), so the runner is arm64 now. The `osx-64` matrix leg installs an x86_64 conda environment, and `conda pack` runs that environment's x86_64 Python, which means it executes under Rosetta on an arm64 host. The `unable to mmap __TEXT` abort looks like a Rosetta translation flake. The two other errors on the same job, the runner's own `System.IO.IOException: Illegal byte sequence` in `PagingLogger`, and the `Can't find action.yml` on the post-run checkout cleanup, both happen after that abort, which also seems to be why the job logs come back empty (`BlobNotFound`) when you try to read them.

It is intermittent: the same step passes on other runs, and only the `osx-64` leg is affected. The `osx-arm64` leg builds an arm64 environment so its Python runs natively and does not hit this.

One option that would remove the Rosetta path entirely is to pin just the `osx-64` row to a native Intel runner. GitHub still offers `macos-26-intel` (and `macos-15-intel`) as x86_64 images; there is no `macos-latest-intel`, since `-latest` stays arm64. `macos-26-intel` is the newest, so it tracks macOS 26 like the arm64 legs. Something like:

```yaml
- { platform: osx-64, platform_name: macOS x64, os: macos-26-intel, build_platform: osx-64, conda_platform: osx-64 }
```

leaving the arm64 leg on `macos-latest`. Worth noting x86_64 macOS runners go away in Fall 2027, so the osx-64 installer will need a longer term plan around then regardless (cross-arch pack, or dropping the Intel build). A short term alternative is just re-running the failed job, since it is flaky.

Happy to send a PR pinning the `osx-64` leg to `macos-15-intel` if that direction sounds right.

Failing run for reference: https://github.com/jupyterlab/jupyterlab-desktop/actions/runs/28604550262

I traced the log chain with Claude Code and confirmed the runner image move and the Intel runner timelines myself before writing this up.

notluquis:
Closes #1018.

The `macOS x64 installer` job fails intermittently at `conda pack` with `rosetta error: unable to mmap __TEXT` and `Abort trap: 6` (exit 134). `macos-latest` is arm64 now, so the `osx-64` leg installs an x86_64 conda environment and `conda pack` runs that environment's x86_64 Python under Rosetta on the arm64 host, which is where the abort comes from. The `osx-arm64` leg is not affected because its environment is native.

This pins just the `osx-64` row to `macos-26-intel`, the newest x86_64 image GitHub offers. There is no `macos-latest-intel` label (the `-latest` alias stays arm64), and `macos-26-intel` is newer than `macos-15-intel`, so it tracks macOS 26 like the arm64 legs and has the longest runway. The build then runs natively with no emulation. Nothing else in the matrix changes.

Worth noting x86_64 macOS runners retire in Fall 2027, so the osx-64 installer will need a longer term plan around then (cross-arch pack, or dropping the Intel build), but that is separate from unblocking CI now.

I traced the failing run and checked the available Intel labels and their support window with Claude Code before writing this.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,8 +36,9 @@ jobs:
         - { platform: linux-64, platform_name: Linux x64,  os: ubuntu-latest, build_platform: linux-64, conda_platform: linux-64 }
         # Native Intel runner: macos-latest is arm64, and packing the osx-64
         # env there runs its x86_64 Python under Rosetta, which aborts conda
-        # pack intermittently (#1018). macos-15-intel is the last x86_64 image.
-        - { platform: osx-64, platform_name: macOS x64, os: macos-15-intel, build_platform: osx-64, conda_platform: osx-64 }
+        # pack intermittently (#1018). macos-26-intel is the newest x86_64
+        # image (GitHub offers no macos-latest-intel; -latest is arm64 only).
+        - { platform: osx-64, platform_name: macOS x64, os: macos-26-intel, build_platform: osx-64, conda_platform: osx-64 }
         - { platform: osx-arm64, platform_name: macOS arm64, os: macos-latest, build_platform: osx-64, conda_platform: osx-arm64 }
         - { platform: win-64, platform_name: Windows x64, os: windows-latest, build_platform: win-64, conda_platform: win-64 }
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -34,7 +34,10 @@ jobs:
       matrix:
         cfg:
         - { platform: linux-64, platform_name: Linux x64,  os: ubuntu-latest, build_platform: linux-64, conda_platform: linux-64 }
-        - { platform: osx-64, platform_name: macOS x64, os: macos-latest, build_platform: osx-64, conda_platform: osx-64 }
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thanks!

PR #1028, opened 2026-07-07
Harden the news feed against malformed feeds and empty results

### LINKED ISSUE #1025: News feed robustness
Two robustness gaps in the news feed path, both introduced or exposed by the `newsfeed.ts` extraction. Corroborated against master at e050646.

- [ ] On a successful fetch that yields zero parseable items, `_updateNewsList` (`welcomeview.ts` around line 745) still runs `appData.newsList = [...newsList]; appData.save()` before the `newsList.length > 0` check. So if the feed endpoint returns a maintenance page or the feed schema changes, `parseNewsFeed` returns an empty array (it no longer throws the way the old inline loop did), and the empty list overwrites and persists over the cached offline items. The previous behavior left the cache intact on a bad fetch. Persisting only when the parsed list is non-empty restores that.

- [ ] `parseNewsFeed` (`newsfeed.ts` around line 21) takes `item.title` and `item.link` straight from the parser output. A feed whose `<title>` or `<link>` comes through as an object (because it carries attributes or nested nodes) or is absent ends up rendering as `[object Object]` or the literal string `undefined`. Coercing to a string and skipping items without a usable title/link keeps a malformed feed from producing garbage rows.

notluquis:
Closes #1025.

Two small robustness fixes in the news feed path, both surfaced while going over the changes since v4.2.5-1 for the release.

`parseNewsFeed` took `item.title` and `item.link` straight from the parser. A feed whose `<title>` comes through as an object (nested nodes or attributes) or whose `<link>` is missing rendered as `[object Object]` or the literal string `undefined`. It now coerces both to a string and skips an item that ends up without a usable title or link, so the `INewsItem` string contract holds even for a malformed feed.

`_updateNewsList` persisted `appData.newsList` on every successful fetch before checking the result was non-empty. Now that `parseNewsFeed` returns an empty list instead of throwing, a maintenance page or a feed schema change would overwrite the cached offline news on disk and blank the panel on the next cold start. Persisting only a non-empty list keeps the cache intact on a bad fetch, matching the pre-refactor behavior.

I extended `test/unit/newsfeed.test.ts` with three cases (missing link, missing title, non-string title). They fail against the current `parseNewsFeed` and pass after the change; the full unit suite (452) stays green and tsc is clean.

I worked through this with Claude Code and verified each step locally before opening the PR.

## AI usage


krassowski (approved):
Thanks!

PR #1037, opened 2026-07-07
Fix env-select popup placement on narrow windows and dismissal on blur

### LINKED ISSUE #1033: Env-select popup placement and dismissal
Two placement/dismissal issues with the Python-environment selector popup, verified against master (e050646).

- [ ] `_resizeEnvSelectPopup` (around line 1368) positions the fixed 600px popup at `x = titleBarRect.width - paddingRight - 600` with no clamp. The window minWidth is 400, so below about 730px (win/linux) or 633px (mac) x goes negative and the popup is clipped off the left edge, leaving its search box unreachable. Clamp x to at least 0, or re-anchor to the window.
- [ ] The popup is not dismissed when the window loses focus: the `blur` handler only deactivates the title bar, and nothing hides the popup. Alt-Tab away and it lingers top-most; on return the refocus path can also close it out from under the user. Hide the popup on window blur.

## Reproduce

Off-screen popup (F4):

1. Start a local session so the environment selector button appears in the title bar.
2. Resize the window narrow, toward its 400px minimum (or start on a small display).
3. Click the environment/server button in the title bar to open the environment picker.
4. See the 600px popup positioned off the left edge, its search box clipped and unreachable.

Lingering popup (F9): open the popup, then Alt-Tab to another application and back without clicking the lab content. The popup is still there on top.

## Expected behavior

The popup stays within the window bounds (its x is clamped so the search box is reachable), and it is dismissed when the window loses focus.

## Context

- Operating System and version: all (F4 is most visible on narrow windows)
- JupyterLab-Desktop version: master (e050646)

## Screenshots (before)

**Env-select popup on a narrow window (F4):** the 600px popup is clipped off the edge and its content is unreachable.

Image

notluquis:
Closes #1033.

Two small fixes for the environment select popup.

The popup is 600px wide and its left edge was computed as the title bar width minus the popup width and a platform padding. The window minimum width is 400px, so on a narrow window that value goes negative and the popup renders clipped off the left edge with its search box unreachable. The left coordinate is now clamped to zero so the popup stays on screen; the width and vertical position are unchanged.

The popup also stayed open when the window lost OS focus, because the blur handler only deactivated the title bar. Since the popup is a child view of the same window, focusing it does not blur the window, so a blur fires only on a genuine focus loss such as Alt-Tab or switching apps, which is exactly when the popup should close. The blur handler now hides it, and the hide path already early-returns when nothing is open.

A unit test drives the real resize method with a 400px title bar and asserts the resulting x is non-negative. It fails before the clamp (x resolves to -233) and passes after. I did not add a unit test for the blur behavior because the handler is wired up inside load() as a closure over local view state, and exercising it cleanly would require standing up the whole window, which felt heavier than the fix warrants.

I built and verified this with Claude Code. Typecheck is clean and the full unit suite is green at 450 tests.

## Before / after

**Env-select popup on a narrow window (F4):**

| Before (master) | After (this PR) |
| --- | --- |
| F4-before | F4-after  |

### DIFF
--- a/src/main/sessionwindow/sessionwindow.ts
+++ b/src/main/sessionwindow/sessionwindow.ts
@@ -182,6 +182,7 @@ export class SessionWindow implements IDisposable {
     });
     this._window.on('blur', () => {
       titleBarView.deactivate();
+      this._hideEnvSelectPopup();
     });
 
     titleBarView.load();
@@ -1365,7 +1366,10 @@ export class SessionWindow implements IDisposable {
     );
 
     this._envSelectPopup.view.view.setBounds({
-      x: Math.round(titleBarRect.width - paddingRight - popupWidth),
+      x: Math.max(
+        0,
+        Math.round(titleBarRect.width - paddingRight - popupWidth)
+      ),
       y: Math.round(titleBarRect.height),
       width: popupWidth,
       height: Math.round(maxHeight)
--- a/src/main/sessionwindow/sessionwindow.ts
+++ b/src/main/sessionwindow/sessionwindow.ts
@@ -1357,8 +1357,13 @@ export class SessionWindow implements IDisposable {
     }
 
     const titleBarRect = this._titleBarView.view.getBounds();
-    const popupWidth = 600;
     const paddingRight = process.platform === 'darwin' ? 33 : 127;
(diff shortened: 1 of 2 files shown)


krassowski (approved):
Nice, thanks!

PR #1038, opened 2026-07-07
Fix title bar maximize/restore desync and cross-window focus theft

### LINKED ISSUE #1034: Title bar maximize state and view focus
Title-bar state and keyboard-focus issues from the WebContentsView migration, verified against master (e050646).

- [ ] The custom title bar maximize/restore buttons toggle only client-side inside their own click handlers; `window.on('maximize'/'unmaximize'/'restore')` only resize and never tell the TitleBarView to swap the buttons. Maximizing by double-click or an OS shortcut leaves the wrong button shown, so the next click is a dead click. Confirmed with a probe on a built app: after a programmatic `maximize()` the window is maximized but the title bar still shows the maximize button. Drive the button state from an IPC on those window events using `win.isMaximized()`, and drop the client-side toggle.
- [ ] The window/title-bar focus handlers only focus `_labView`; in Welcome mode it is null, so the welcome view never receives keyboard focus on window activation. Focus the active content view instead.
- [ ] The labView `did-finish-load` handler focuses its own webContents unconditionally, so a background second session window steals focus when its lab reloads. Guard on the window being focused.

Related, needs per-OS testing on a packaged build (cannot be automated, and app-region behavior in a child WebContentsView is a known Electron soft spot, e.g. electron/electron#43320): confirm the main window can still be dragged by its title strip and that the title-bar buttons are clickable on macOS, Windows and Linux. The title bar is rendered inside a child WebContentsView, and the JS mousedown fallback does not move the window, so if app-region drag regressed the window would be undraggable. If it has, the fix is to make the fallback perform a real window move.

## Reproduce

Maximize/restore desync (F5), clearest on Windows and Linux where the custom min/max/close control box is the primary window chrome (the state desync also reproduces on macOS at the DOM level):

1. Open a window.
2. Maximize it by double-clicking the title-bar drag strip, or with the OS maximize shortcut or window snap, not the custom maximize button.
3. See the title bar still showing the Maximize button instead of Restore.
4. Click that button: the first click is a no-op on the already-maximized window and only then swaps the icon.

Empirical check on a built app: after a programmatic `win.maximize()` the window reports `isMaximized() === true`, but the title bar still shows the maximize button (restore stays hidden).

Welcome focus (F10): on the Welcome screen (no session), Alt-Tab back to the window and try to Tab through the welcome UI; keyboard focus is not in the welcome content until you click it.

Cross-window focus (F11): open two session windows, work in A, and let B's lab reload in the background; focus jumps to B.

## Expected behavior

The maximize/restore button always reflects the real window state; the active content view receives keyboard focus on activation; a background window does not steal focus on reload.

## Context

- Operating System and version: F5 most visible on Windows/Linux; F10/F11 all
- JupyterLab-Desktop version: master (e050646)

notluquis:
Closes #1034.

Three focused fixes for the custom title bar and window focus handling:

- Maximize/restore button desync. The buttons only toggled themselves inside their own onclick handlers, so maximizing via a double click on the drag region or via the OS window controls left the wrong button showing. The button state now follows a new `SetMaximized` renderer event driven by the window's own `maximize`/`unmaximize`/`restore` events, using the same IPC path as `SetActive` and `ShowServerStatus`, so the real window state is the single source of truth.

- Welcome view never keyboard-focused. The window and titlebar focus handlers only forwarded focus to the lab view, which is null in Welcome mode, so the welcome content never got focus. They now focus the active content view with a destroyed-guard, which covers both Welcome and Lab.

- Background window steals focus. The lab view's `did-finish-load` handler focused itself unconditionally, so a background second window grabbed focus whenever its lab reloaded. It now only focuses when its own window is already focused.

The focus-handler change touches the same handlers as #1031, so a trivial rebase may be needed at merge time.

Added unit coverage for `TitleBarView.setMaximized` and the new `onSetMaximized` preload bridge. The focus-handler changes are closures registered during window construction, so I did not add tests that would need the whole window wiring stood up rather than asserting real behavior.

I built and verified this locally with Claude Code: `tsc --noEmit` is clean and the unit suite passes (452 tests).

## Verification, before and after

Ran an automated check on a built app (macOS): launch, then `BrowserWindow.maximize()` (the OS or double-click path, not the custom button), then read the title bar button state.

Before this change (master build): `isMaximized() === true`, but the title bar still shows the maximize button (`#maximize-button` display `block`, `#restore-button` display `none`).

After this change: `isMaximized() === true`, and the title bar shows restore (`#maximize-button` display `none`, `#restore-button` display `block`).

Caveat on platforms: the custom min/max/close control box is zero-width on macOS (the native traffic lights are used there), so the check above is a DOM-level before/after, not a visible one. I could not test this on Windows or Linux, where that control box is the primary window chrome and the swap is actually visible, so that part is expected from the code and the macOS DOM state rather than verified. Someone on Windows or Linux should confirm it and, ideally, attach a short before/after recording.

## Follow-up tests for the drag region and focus fixes

Added `test/e2e/titlebar-dragregion.test.ts`, a regression guard for the window drag region. It launches the app, finds the title bar page by its `#app-title` element, and asserts the computed `webkitAppRegion` is `drag` on the title and `no-drag` on the close and server buttons. I confirmed it is a real guard by removing the `-webkit-app-region: drag` rule from the built title bar: the computed value falls back to `none` and the test fails. It is a CSS-declaration guard only. An actual OS window move cannot be driven from Playwright (synthetic mouse events do not fire the app-region hit-test), so the real drag and the layered-WebContentsView no-drag hit-testing (electron/electron#43320) still need a manual pass per OS.

I looked into automating the two focus fixes as well but could not make either deterministic. Under the Playwright/Electron harness the launched app never becomes the OS-frontmost application, so `webContents.isFocused()` reads `false` for every view even after forcing `BrowserWindow.focus()` and `webContents.focus()` from the main process. Any assertion built on that would pass or fail regardless of the fix, so I left them as manual steps rather than ship a flaky or tautological test.

Manual verification for the focus fixes:

- Welcome view focus (Welcome mode): start the app so the Welcome view is showing, click another app to defocus, then click the JupyterLab window title bar to refocus it. Press Tab. Focus should land inside the welcome content (a recent-session entry or a new-notebook link highlights). Before the fix, Tab did nothing because focus was forwarded to the null lab view.
- Background window does not steal focus: open two session windows, keep window A in the foreground, and trigger a reload of the lab in background window B (for example switch its environment so its lab view reloads). Focus should stay on window A. Before the fix, B grabbed focus on its `did-finish-load`.

### DIFF
--- a/src/main/eventtypes.ts
+++ b/src/main/eventtypes.ts
@@ -96,6 +96,7 @@ export enum EventTypeRenderer {
   SetRunningServerList = 'set-running-server-list',
   SetTitle = 'set-title',
   SetActive = 'set-active',
+  SetMaximized = 'set-maximized',
   ShowServerStatus = 'show-server-status',
   ShowServerNotificationBadge = 'show-server-notification-badge',
   SetRecentSessionList = 'set-recent-session-list',
--- a/src/main/sessionwindow/sessionwindow.ts
+++ b/src/main/sessionwindow/sessionwindow.ts
@@ -192,10 +192,16 @@ export class SessionWindow implements IDisposable {
     // this._labView freshly means an env switch that recreates the labView cannot
     // leave a stale, disposed webContents behind, nor accumulate a handler per switch.
     this._window.webContents.on('focus', () => {
-      this._labView?.view?.webContents?.focus();
+      const wc = this.contentView?.webContents;
+      if (wc && !wc.isDestroyed()) {
+        wc.focus();
+      }
     });
     this._titleBarView.view.webContents.on('focus', () => {
-      this._labView?.view?.webContents?.focus();
+      const wc = this.contentView?.webContents;
+      if (wc && !wc.isDestroyed()) {
+        wc.focus();
+      }
     });
 
     if (this._contentViewType === ContentViewType.Lab) {
@@ -240,12 +246,15 @@ export class SessionWindow implements IDisposable {
       this._resizeViewsDelayed();
     });
     this._window.on('maximize', () => {
+      this._titleBarView.setMaximized(this._window.isMaximized());
       this._resizeViewsDelayed();
     });
     this._window.on('unmaximize', () => {
+      this._titleBarView.setMaximized(this._window.isMaximized());
       this._resizeViewsDelayed();
     });
     this._window.on('restore', () => {
+      this._titleBarView.setMaximized(this._window.isMaximized());
       this._resizeViewsDelayed();
     });
     this._window.on('move', () => {
@@ -374,7 +383,9 @@ export class SessionWindow implements IDisposable {
     this._window.contentView.addChildView(labView.view);
 
     labView.view.webContents.on('did-finish-load', () => {
-      labView.view.webContents.focus();
+      if (this._window.isFocused()) {
+        labView.view.webContents.focus();
+      }
     });
 
     labView.load((errorCode: number, errorDescription: string) => {
--- a/src/main/titlebarview/preload.ts
+++ b/src/main/titlebarview/preload.ts
@@ -5,11 +5,13 @@ const { contextBridge, ipcRenderer } = electron;
 
 type SetTitleListener = (title: string) => void;
 type SetActiveListener = (active: boolean) => void;
+type SetMaximizedListener = (isMaximized: boolean) => void;
 type ShowServerStatusListener = (show: boolean) => void;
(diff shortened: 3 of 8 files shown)


krassowski (approved):
LGTM, let's do a quick copilot pass

krassowski:
^ it sounds like edge-cases possibly worth follow-up later but not blocking.

PR #1067, opened 2026-07-22
Change release PR environment

krassowski:
Follow-up to #1066


krassowski:
This is largely the same as https://github.com/jupyterlab/jupyterlab-desktop/pull/1017

PR #1071, opened 2026-07-26
docs: correct the minimum supported OS versions

notluquis:
The install table says macOS 10.15+, but that has not been true for a while. `mac.minimumSystemVersion` is `12.0.0` in the build config, and Electron 42 lists Monterey as its floor, so someone on 10.15 or 11 downloads a `.dmg` that will not launch. This corrects it to 12+.

While there, the Linux column had no versions at all, so I filled in the ones Electron verifies its prebuilt binaries against (Ubuntu 18.04+, Fedora 32+, Debian 10+). Windows 10/11 was already right.

One line, no behavior change. I took the numbers from [Electron 42's platform support section](https://github.com/electron/electron/blob/v42.0.0/README.md) and our own `build.mac.minimumSystemVersion` rather than guessing; worked through it with Claude Code.


krassowski (approved):
Thanks!

PR #1078, opened 2026-07-26
Refresh vulnerable transitive deps within their declared ranges

notluquis:
Twenty-one of the open Dependabot alerts are transitive build-time deps. For all of them the patched version already satisfies the range their parent declares, so the lockfile was just holding older resolutions and nothing was actually blocking the fix.

Dropping those entries and reinstalling moves each to a patched release:

| package | before | after | fixed in |
| --- | --- | --- | --- |
| `@xmldom/xmldom` | 0.8.10 | **0.8.13** | 0.8.13 |
| `brace-expansion` | 1.1.11 / 2.0.1 / 2.1.1 / 5.0.6 | **1.1.16 / 2.1.2 / 5.0.8** | 1.1.16 / 2.1.2 / 5.0.7 |
| `fast-uri` | 3.1.2 | **3.1.4** | 3.1.4 |
| `js-yaml` | 4.1.0 / 5.2.0 | **4.3.0 / 5.2.2** | 4.3.0 |
| `minimatch` | 3.1.2 / 5.1.6 | **3.1.5 / 5.1.9** | 3.1.3 / 5.1.8 |
| `picomatch` | 2.3.1 | **2.3.2** | 2.3.2 |
| `tar` | 7.5.16 | **7.5.22** | 7.5.18 |
| `tmp` | 0.2.1 | **0.2.7** | 0.2.6 |
| `vite` | 8.0.11 | **8.1.5** | 8.0.16 |

`package.json` is untouched. No `resolutions`, no parent bumps: every one is a patch or minor inside the major the parent already asked for.

### Why this is not the "bump the parent" case

Four transitive Dependabot PRs were closed earlier (#956 xmldom, #957 postcss, #923 tmp, #864 micromatch) on the reasoning that bumping the parent would carry them along. For these that does not hold, and I should own that since I closed them.

`@xmldom/xmldom` is the clearest example. It arrives through `electron-builder` → `app-builder-lib` → `plist@3.1.0`, which declares `@xmldom/xmldom: ^0.8.8`. I checked `app-builder-lib` at both 26.15.3 and 26.15.7 and the declared dep is identical, so no electron-builder bump would ever have moved it. The patched 0.8.13 sits inside `^0.8.8` and only needed the lockfile to re-resolve. `postcss` lands here for the same reason.

### The rest of the diff

35 packages moved in total. Nine are the targets above; the remaining collateral is vite's own tree following it (`rolldown`, the `napi`/`emnapi` bindings, `postcss`, `nanoid`) plus `rimraf@3.0.2` dropping out with no remaining parent. All dev tooling, and none of it ships in the asar since yarn 1 omits devDependencies from the package.

I audited the diff package by package rather than trusting the install; the list above is that audit, not a summary of the install output.

### Checks

`yarn type-check`, `yarn test:unit` (469) and a full `yarn build` are green on the new lockfile.

Worked through this with Claude Code.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -148,25 +148,25 @@
     minimatch "^9.0.3"
     plist "^3.1.0"
 
-"@emnapi/core@1.10.0":
-  version "1.10.0"
-  resolved "https://registry.yarnpkg.com/@emnapi/core/-/core-1.10.0.tgz#380ccc8f2412ea22d1d972df7f8ee23a3b9c7467"
-  integrity sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==
+"@emnapi/core@1.11.1":
+  version "1.11.1"
+  resolved "https://registry.yarnpkg.com/@emnapi/core/-/core-1.11.1.tgz#b9e1064f3a6b1631e241e638eb48d736bfd372a6"
+  integrity sha512-RSvbQmHzdKzNsLYa/wHrbc3KN4sYLKAdPZxqiM2HATqv/SBk2/ENSHpvXGaLOMcsAyz0poEGqkmmKYG3OWiJEQ==
   dependencies:
-    "@emnapi/wasi-threads" "1.2.1"
+    "@emnapi/wasi-threads" "1.2.2"
     tslib "^2.4.0"
 
-"@emnapi/runtime@1.10.0":
-  version "1.10.0"
-  resolved "https://registry.yarnpkg.com/@emnapi/runtime/-/runtime-1.10.0.tgz#4b260c0d3534204e98c6110b8db1a987d26ec87c"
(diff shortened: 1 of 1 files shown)


notluquis:
Closing this. It was the wrong shape: Dependabot already covers almost all of it, and hand-editing the lockfile just makes six reviewable per-package PRs into one opaque 412-line diff and leaves those PRs stale.

Overlap with what is already open: `tmp` #969, `vite` #985, `js-yaml` #1058, `fast-uri` #1061, `tar` #1062. And `@xmldom/xmldom` 0.8.10 to 0.8.13 is #956, which I closed earlier on the reasoning that bumping the parent would carry it. That reasoning was wrong, `plist` still declares `^0.8.8` today so no parent bump moves it, and the patched 0.8.13 sits inside that range. Asking Dependabot to reopen it instead.

The one thing worth keeping from this: `brace-expansion`, `minimatch` and `picomatch` have no PR of their own. That looks like `open-pull-requests-limit: 10` in `.github/dependabot.yml` against 17 currently open, so the queue is saturated rather than Dependabot being unable to fix them. They should follow eslint and vitest in #1060 anyway. Merging the backlog is what unblocks the rest.

PR #1091, opened 2026-07-27
Snap on stable cannot launch: refresh electron-builder past the .tar.7z extraction bug

### LINKED ISSUE #1085: Snap fails to launch after kernel upgrade to 7.0.0 on Ubuntu 24.04: "desktop-init.sh: No such file or directory"
## Reproduce

2. Click on '...'
3. Scroll down to '...'
4. See error '...'

## Context

- Operating System and version: 
- JupyterLab-Desktop version: 
- Installer: 

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the terminal inside of JupyterLab Desktop here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Browser Output (required for interface issues after loading)</summary>

<pre>
Paste the output from the DevTools JavaScript console here, if applicable.

</pre>
</details>

<details><summary>Logs (required for issues with loading, server and startup)</summary>

<pre>
Paste the logs from the `main.log` file here (optionally also from `renderer.log`).

</pre>
</details>

notluquis:
Fixes #1085. The snap on the store cannot launch, so this one is time sensitive.

## What is wrong

The 4.6.2-1 snap has the template tarball sitting at its root instead of the files inside it:

```
squashfs-root/command.sh
squashfs-root/jupyterlab-desktop
squashfs-root/snap-template-electron-4.0-2-amd64.tar     <- never unpacked
```

`command.sh` starts with `exec "$SNAP/desktop-init.sh" ...`, and that script is one of the three inside the tarball, so every launch ends at:

```
/snap/jupyterlab-desktop/35/command.sh: line 2:
/snap/jupyterlab-desktop/35/desktop-init.sh: No such file or directory
```

The Snap Store API reports revision 35 (4.6.2-1) on both `stable` and `candidate` for amd64, released 2026-07-22. `edge` still has revision 31 (4.2.5-1), which works. The deb, rpm and the other platforms are unaffected; this is specific to the snap template path.

## Cause

`extractArchive` in app-builder-lib branched on extension and, for a `.tar.7z` artifact, stripped only the outer 7z layer, leaving the inner tar untouched. The snap template is published as `snap-template-electron-4.0-2-amd64.tar.7z`, so `mksquashfs` packed the tar verbatim. That is [electron-userland/electron-builder#10002](https://github.com/electron-userland/electron-builder/issues/10002), which affects 26.15.0 through 26.15.6 and is fixed in 26.15.7 (2026-07-18). Our lockfile resolved 26.15.3 even though `^26.15.3` already allowed the fix four days before we built the release.

## This PR

- refreshes electron-builder to 26.15.7. Diff is one line of package.json and the lockfile entries for electron-builder, app-builder-lib and dmg-builder.
- adds a five line step: the `$SNAP/...` paths `command.sh` execs, minus the paths present in the image, has to be empty. It reads the requirement off the launcher rather than hardcoding filenames, so a build that stops using the template path, or an upstream rename, does not fail a correct release.

  Canonical's advice for this is `snap install --dangerous` plus [review-tools](https://snapcraft.io/review-tools) in CI. Neither fits: review-tools checks store requirements rather than whether the launcher's own dependencies are in the image, and installing and running the app needs snapd and a display on the runner, which trades this failure for flakier ones. A set difference over the file list is the smallest thing that catches the class.

The second half matters as much as the first. This workflow runs on pull requests too, so the check would have failed on the PR that introduced the breakage rather than after it reached the store, and the snap is the one artifact nothing ever looks inside. The upstream fix also renames the toolset cache directory for `.tar.7z`, so a poisoned cache on a builder cannot survive the upgrade.

## What still needs doing after this lands

A rebuilt snap has to be published and promoted, since revision 35 stays on stable until a new revision replaces it. That is a maintainer action, I cannot trigger it.

## Verified

Unpacked the released artifact and confirmed the missing scripts and the stray tarball, queried the Snap Store for the published revisions, and checked that the `.tar.7z` branch is present in 26.15.7 and absent in 26.15.3 through 26.15.6. Build and the unit suite pass on the refreshed lockfile.

The check itself I ran both ways against the released 4.6.2-1 snap: it names the three missing scripts and exits 1, and with those entries present it exits 0. What I could not do is build a snap on macOS, so its first run against a freshly built artifact happens in CI here.

Worked through this with Claude Code, and the artifact inspection and version bisect are mine.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -197,18 +197,12 @@ jobs:
           command -v unsquashfs >/dev/null || sudo apt-get install -y --no-install-recommends squashfs-tools
           rm -rf snap-check
           unsquashfs -q -d snap-check dist/JupyterLab.snap command.sh
-          unsquashfs -l dist/JupyterLab.snap > snap-contents.txt
-          status=0
-          # whatever the launcher execs has to be in the image: the 4.6.2-1 snap
-          # shipped the template as a tarball and died on a missing script
-          for target in $(grep -oE '\$SNAP/[A-Za-z0-9._/-]+' snap-check/command.sh | sort -u); do
-            entry="squashfs-root/${target#\$SNAP/}"
-            if ! grep -qx "$entry" snap-contents.txt; then
-              echo "::error::command.sh execs $target but it is not in the snap"
-              status=1
-            fi
-          done
-          exit $status
+          unsquashfs -l dist/JupyterLab.snap | sort > snap-contents.txt
+          # 4.6.2-1 shipped the template as a tarball, so the launcher died on a
+          # missing script: whatever command.sh execs has to be in the image
+          grep -oE '\$SNAP/[^" ]+' snap-check/command.sh | sed 's|\$SNAP|squashfs-root|' | sort -u > snap-needs.txt
+          comm -23 snap-needs.txt snap-contents.txt | tee snap-missing.txt
+          [ ! -s snap-missing.txt ]
 
       - name: Upload Snap Installer
         if: matrix.cfg.platform == 'linux-64'
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -194,21 +194,21 @@ jobs:
       - name: Verify the snap can actually launch
         if: matrix.cfg.platform == 'linux-64'
         run: |
-          sudo apt-get update -qq
-          sudo apt-get install -y --no-install-recommends squashfs-tools
+          command -v unsquashfs >/dev/null || sudo apt-get install -y --no-install-recommends squashfs-tools
+          rm -rf snap-check
+          unsquashfs -q -d snap-check dist/JupyterLab.snap command.sh
           unsquashfs -l dist/JupyterLab.snap > snap-contents.txt
-          missing=0
-          for script in desktop-init.sh desktop-common.sh desktop-gnome-specific.sh; do
-            if ! grep -qx "squashfs-root/$script" snap-contents.txt; then
-              echo "::error::$script is missing from the snap; command.sh execs it, so the app cannot start"
-              missing=1
+          status=0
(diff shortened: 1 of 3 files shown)


notluquis:
@krassowski bumping this one, since `stable` is serving a snap that cannot start and stays that way until a rebuilt revision replaces it.

.github/workflows/publish.yml
```diff
@@ -200,6 +200,28 @@ jobs:
...
+          fi
+          missing="$(comm -23 "$RUNNER_TEMP/snap-need.txt" "$RUNNER_TEMP/snap-have.txt")"
+          if [ -n "$missing" ]; then
+            echo "::error::command.sh execs paths that are not in the snap: $missing"
+            exit 1
+          fi
+          if grep -qE '^squashfs-root/snap-template-.*\.tar$' "$RUNNER_TEMP/snap-have.txt"; then
+            echo "::error::the snap template was packed as a tarball instead of being extracted"
+            exit 1
+          fi
```
krassowski, line 223: This does not verify that snap can launch, it only tests for this specific bug, though I don't know how much better we could do with snapcraft

krassowski (approved):
Thanks!

.github/workflows/publish.yml
notluquis: Agreed, it infers rather than verifies. #1094 replaces it with installing the built snap and running it: `snap install --dangerous --classic` (what Canonical documents for a local build) and then `jupyterlab-desktop --version` under xvfb, which quits before any window is created, so the whole launcher chain executes. Same shape as jupyter_releaser's check_dist, which installs a built Python dist into a venv and runs it rather than only checking it. The open question is whether snapd works on a GitHub-hosted runner at all, since it is not in the image manifest, and that PR's first run answers it.

PR #1092, opened 2026-07-27
Refuse browser permissions to content that is not the Jupyter server and limit permissions grants

### LINKED ISSUE #1089: No session sets a permission request handler, so embedded content is granted permissions silently
No session in the app calls `setPermissionRequestHandler`, so Electron's default applies to every view.

Measured with a small harness (a page loaded in a `WebContentsView`, once on `session.defaultSession` and once on a named partition, with and without a handler):

| | `Notification.requestPermission()` | geolocation |
| --- | --- | --- |
| no handler (what we ship) | `granted`, no prompt | not denied |
| deny-all handler | `denied` | `code 1 User denied Geolocation` |

This is reachable, not theoretical: a server-initiated redirect can render off-origin content inside the lab view today, and after #1087 the sign-in window renders provider pages on the same session by design. Electronegativity flags the absence as [`PERMISSION_REQUEST_HANDLER_JS_CHECK`](https://github.com/doyensec/electronegativity/wiki/PERMISSION_REQUEST_HANDLER_JS_CHECK).

## What JupyterLab actually needs

Grepping the shipped JupyterLab bundle in a current environment: `navigator.clipboard` appears 5 times, `Notification.requestPermission` and `getUserMedia` do not appear at all. So a deny-by-default handler with a small allowlist for the server origin should not break core; extensions that need more can be handled when a concrete one shows up.

## Proposal

- one handler per session, deny by default, allow a short list only when the requesting URL is on the Jupyter server origin
- `setPermissionCheckHandler` alongside it, since the synchronous checks bypass the request handler
- log denials at debug level so a missing permission is diagnosable rather than mysterious

Found while working on #1083. Numbers above are from my own runs, done with Claude Code.

notluquis:
Closes #1089.

No session in the app calls `setPermissionRequestHandler`, so Electron's default applies to every view: `Notification.requestPermission()` returns `granted` with no prompt, and geolocation is not denied. Electronegativity flags the absence as [`PERMISSION_REQUEST_HANDLER_JS_CHECK`](https://github.com/doyensec/electronegativity/wiki/PERMISSION_REQUEST_HANDLER_JS_CHECK).

Reachable today: a server-initiated redirect renders an off-origin page inside the lab view, and after #1087 the sign-in window renders provider pages on the same session by design.

`isPermissionAllowed({ permission, requestingUrl, serverUrl })` grants only when the request comes from the Jupyter server origin of the window that owns the sender, and only for `clipboard-read`, `clipboard-sanitized-write`, `fullscreen` and `notifications`. Everything else is refused for every origin.

Measured against a real Electron session, with a page on one origin and a cross-origin iframe inside it:

| request | from | verdict |
| --- | --- | --- |
| `notifications`, main frame | server origin | granted |
| `clipboard-read`, main frame | server origin | granted |
| `geolocation`, main frame | server origin | denied |
| `notifications`, subframe | a different origin | denied |

The subframe row is why the request handler reads `details.requestingUrl` and never falls back to `webContents.getURL()`: the fallback is the top frame's URL, so an off-origin iframe would be measured against the server origin and granted on that basis. Electron documents `requestingUrl` as the URL of the requesting frame, and the run above confirms it reports the iframe's own origin. The check handler uses `requestingOrigin` for the same reason, since `requestingUrl` is empty there for several permissions.

Two details that decide where the code lives: sessions are hooked through `app.on('session-created')` so a remote partition is covered as it appears, and the origin is resolved per `webContents` rather than captured once, because local sessions share the default session on different ports.

**The trade worth stating.** Same-origin includes anything the Jupyter server serves, and a notebook's HTML output is served from that origin. So `clipboard-read` is granted to code running inside any notebook the user opens, not only to JupyterLab itself, which means a notebook can read the system clipboard while the window is focused. JupyterLab does need it: `packages/apputils/src/clipboard.ts` calls `systemClipboard.readText()` and the terminal extension reads the clipboard to paste. Dropping it would break paste. Naming it here so it is a decision rather than something found later.

The other behaviour change worth arguing about: `media` is refused even on the server origin, so a notebook cannot reach the camera or microphone. For a local server that is arguably overcautious, since the kernel already runs arbitrary code on the same machine; for a remote one it is the difference between a page served by someone else's machine and your webcam. VS Code allows `media` for its own window and denies it for webview content; Signal defaults it off behind a user setting. Passing `isRemote` into the predicate, or gating on a setting, are both a few lines if either is preferred.

The obvious candidate for something this breaks does not apply: `ipywebrtc` is the only Jupyter package I found using `getUserMedia`, and its peer range tops out at `@jupyter-widgets/base ^4.0` with a `webpack.config.lab3.js` build, so it cannot load against the ipywidgets 8.1.8 this app bundles. JupyterLab core itself reaches for `requestFullscreen` and the clipboard and nothing else: no `getUserMedia`, no `requestPermission`, no `geolocation`.

`pointerLock` is deliberately absent. VS Code grants it unconditionally, but that is for its integrated browser where any site runs; `OrbitControls` in three.js does not use it, and neither pythreejs, ipyvolume nor K3D reference it at all.

Worked through this with Claude Code; the table is from a harness I ran against a real session with a cross-origin iframe, not from the docs.

## AI usage


notluquis:
@krassowski ready for review: checks green and out of draft. A Copilot review requested from this account comes back at its quota limit, so triggering that one needs an account that has quota. Anything it flags gets fixed or answered with evidence.

src/main/permissionpolicy.ts
```diff
@@ -0,0 +1,40 @@
...
+const SERVER_ORIGIN_PERMISSIONS: Record<string, boolean> = {
+  'clipboard-read': true, // JupyterLab copies and pastes cells through it
+  'clipboard-sanitized-write': true,
+  fullscreen: true, // presentations
+  notifications: true, // extensions report long running work
+  media: false, // a remote server's page must not reach camera or microphone
+  geolocation: false,
+  'display-capture': false,
+  'idle-detection': false
+};
```
krassowski: Might be worth documenting these choices in user-facing docs later.

krassowski (dismissed):
Thanks. This makes sense to me. Local validation with Opus highlights potential issue on `notifications` grant:

> The notifications grant never takes effect, because Electron passes a `null` `webContents` for every notification permission check. The check handler at app.ts:614 resolves the server origin from `webContents`, so `_serverUrlForWebContents(null)` returns undefined and permissionpolicy.ts:36 refuses. `Notification.requestPermission() still` resolves to granted through the request handler, which is the row the harness measured, but `Notification.permission` then reads denied and the display path fires an error event instead of showing anything.
Path in Electron 42: BlinkNotificationServiceImpl calls GetPermissionResultForOriginWithoutContext, which calls GetPermissionStatus, which calls CheckPermissionWithDetails(permission, {}, ...) with an empty frame host. The session docs say the same thing: "certain other permission checks such as notifications checks will always pass null".
Same root cause reaches worker contexts: GetPermissionResultForWorker also passes no frame host. Matching requestingOrigin against the set of server origins known to the open windows, and using webContents only to narrow it when present, would fix both.

If confirmed that could be a mild blocker, but fine to address in a separate PR. Feel free to merge otherwise.

krassowski (approved):
Thanks!

PR #1093, opened 2026-07-27
Make blocked navigation the default for views nobody claimed

### LINKED ISSUE #1088: Navigation guards are per-view exceptions; five surfaces have none
Navigation control in this app is an exception rather than a policy: the lab view guards itself (and #1087 adds the sign-in window, deliberately unguarded because it *is* the unprivileged surface), while everything else is open by default.

| surface | loads | navigation guard |
| --- | --- | --- |
| welcome view | `data:text/html`, renders the remote news feed | none |
| title bar | `data:text/html` | none |
| themed window / themed view (dialogs) | `data:text/html` | none |
| connect window (`connect.ts`) | the remote server URL, navigates freely by design | none |
| lab view | server URL | yes |

None of the `data:` views is supposed to navigate at all. Today nothing stops one, and the welcome view is the one that renders content derived from a remote feed, which is the surface GHSA-vrj4-r4fw-9rfh was about.

## What the tooling and other apps expect

Doyensec's rule for this is literally named [`LIMIT_NAVIGATION_GLOBAL_CHECK`](https://github.com/doyensec/electronegativity/wiki/LIMIT_NAVIGATION_GLOBAL_CHECK), that is, limits at application level rather than per window, and Electron's own checklist items 13 and 14 are written the same way. In shipped apps: Signal denies window creation globally in `web-contents-created` and also blocks `will-attach-webview`; VS Code calls `preventDefault()` on `will-navigate` for every webContents with an explicit exception for its integrated browser views; Bitwarden guards `will-redirect` alongside `will-navigate`.

## Proposal

```ts
guardWebContents(contents, { policy: 'app-owned' })                       // welcome, title bar, dialogs
guardWebContents(contents, { policy: 'server-view', serverUrl, onAuthChain })  // lab view, the predicate from #1087
// sign-in and connect windows: registered as unprivileged, navigation free by design
```

plus a fallback so a surface added later is closed rather than open:

```ts
app.on('web-contents-created', (_event, contents) => {
  if (isGuarded(contents)) return;
  contents.on('will-navigate', event => event.preventDefault());
  contents.on('will-attach-webview', event => event.preventDefault());
  contents.setWindowOpenHandler(() => ({ action: 'deny' }));
});
```

## To check before merging

- audit every dialog for anchors with a real `href`; the ones I have seen go through IPC to `shell.openExternal`, but that needs confirming case by case
- confirm `connect.ts` keeps working once it has to register explicitly
- the `data:` views should never navigate, so a violation is worth logging rather than silently preventing

Found while working on #1083. Written with Claude Code; the table above comes from grepping this tree, and the behaviour of the other apps from reading their sources rather than their docs.

notluquis:
Closes #1088.

## What is wrong

Navigation control in this app is an exception rather than a policy. The lab view guards itself; everything else is open by default.

| surface | loads | guard before this PR |
| --- | --- | --- |
| welcome view | `data:text/html`, renders the remote news feed | none |
| title bar | `data:text/html` | none |
| themed window and themed view, which every dialog is built from | `data:text/html` | none |
| session window itself | container window | none |
| connect window | the remote server URL, navigates freely by design | none |
| lab view | server URL | yes |

None of the `data:` views is meant to navigate at all. The welcome view is the one that renders content derived from a remote feed, which is the surface GHSA-vrj4-r4fw-9rfh was about.

## What this does

- `installGlobalNavigationGuard()` denies `will-navigate`, `will-redirect` and window creation for any webContents that has not claimed a policy, and refuses `will-attach-webview` everywhere.
- the check runs at navigation time rather than at creation time. `web-contents-created` fires before a constructor can claim its own view, so deciding at creation would either lock out every view or trust every view.
- `guardAppOwnedView()` is what the bundled views claim: stay on this document, and hand a link the user follows to the system browser through the same `http`/`https`/`mailto` allowlist used elsewhere.
- the lab view and the connect window claim themselves with `markGuarded()` and keep the behaviour they have now, one deciding per origin and the other following a login wherever the server sends it.

A view added later inherits the deny rather than inheriting nothing, which is the actual point.

## Prior art

Doyensec's rule is named [`LIMIT_NAVIGATION_GLOBAL_CHECK`](https://github.com/doyensec/electronegativity/wiki/LIMIT_NAVIGATION_GLOBAL_CHECK), that is, limits at application level rather than per window, and Electron's checklist items 13 and 14 are written the same way. Reading how others do it: Signal denies window creation globally in `web-contents-created` and blocks webview attachment, VS Code calls `preventDefault()` on `will-navigate` for every webContents with an explicit exception for its integrated browser views, and Bitwarden guards `will-redirect` alongside `will-navigate`.

## Verified

Audited the bundled views before touching them, since denying navigation there could have broken something real: every anchor is `href="#"` (ten of them) or `href="javascript:void(0)"` (two), none navigates, and none of those views calls `window.open` or uses `target="_blank"`. Nothing legitimate is taken away.

Unit tests cover both sides of the claim check, the app-owned policy and the scheme allowlist, checked by mutation: making the fallback ignore claims fails the claimed-view test, dropping the allowlist fails three. Build, eslint, prettier and the full unit suite (484) pass, and the app starts with nothing logged as blocked. I smoke tested startup and the welcome view rather than opening every dialog one by one.

Independent of #1087 and #1092, and can land in any order relative to them.

Worked through this with Claude Code.

### DIFF
--- a/src/main/app.ts
+++ b/src/main/app.ts
@@ -29,6 +29,7 @@ import {
   setupJlabCLICommandWithElevatedRights,
   waitForDuration
 } from './utils';
+import { installGlobalNavigationGuard } from './navigationguard';
 import { IServerFactory, JupyterServerFactory } from './server';
 import { connectAndGetServerInfo, IJupyterServerInfo } from './connect';
 import { UpdateDialog } from './updatedialog/updatedialog';
@@ -279,6 +280,7 @@ export class JupyterApplication implements IApplication, IDisposable {
     this._serverFactory.createFreeServer().catch(error => {
       console.error('Failed to create free server', error);
     });
+    installGlobalNavigationGuard();
     this._registerListeners();
 
     if (
--- a/src/main/connect.ts
+++ b/src/main/connect.ts
@@ -3,6 +3,7 @@
 
 import { BrowserWindow, Cookie } from 'electron';
 import { clearSession } from './utils';
+import { markGuarded } from './navigationguard';
 
 export let connectWindow: BrowserWindow;
 
@@ -55,6 +56,8 @@ export async function connectAndGetServerInfo(
     }
 
     const window = new BrowserWindow(browserOptions);
+    // this window exists to follow a login wherever the server sends it
+    markGuarded(window.webContents);
 
     const timeout = options?.timeout || 30000;
 
--- a/src/main/dialog/themedview.ts
+++ b/src/main/dialog/themedview.ts
@@ -5,6 +5,7 @@ import { WebContentsView } from 'electron';
 import * as fs from 'fs';
 import * as path from 'path';
 import { DarkThemeBGColor, LightThemeBGColor } from '../utils';
+import { guardAppOwnedView } from '../navigationguard';
 
 export class ThemedView {
   constructor(options: ThemedView.IOptions) {
@@ -14,6 +15,7 @@ export class ThemedView {
         preload: options.preload || path.join(__dirname, './preload.js')
       }
     });
+    guardAppOwnedView(this._view.webContents);
     this._view.setBackgroundColor(
       this._isDarkTheme ? DarkThemeBGColor : LightThemeBGColor
     );
--- a/src/main/dialog/themedwindow.ts
+++ b/src/main/dialog/themedwindow.ts
@@ -5,6 +5,7 @@ import { BrowserWindow } from 'electron';
 import * as fs from 'fs';
 import * as path from 'path';
 import { DarkThemeBGColor, LightThemeBGColor } from '../utils';
+import { guardAppOwnedView } from '../navigationguard';
 
 export class ThemedWindow {
   constructor(options: ThemedWindow.IOptions) {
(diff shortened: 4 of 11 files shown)


notluquis:
#1087 now builds on this branch: the sign-in window it adds was wiring the same three hooks by hand, so both go through a `guardNavigation(contents, decide)` here and each surface brings its rule instead of a copy of the plumbing. Three callers today: app-owned views hand links to the browser, unclaimed views are refused, the sign-in window follows the chain while it stays on http or https.

This one still stands alone and can merge first; #1087 collapses to its own changes once it does.

notluquis:
@krassowski ready for review: checks green and out of draft. A Copilot review requested from this account comes back at its quota limit, so triggering that one needs an account that has quota. Anything it flags gets fixed or answered with evidence.

src/main/app.ts
```diff
@@ -279,6 +280,7 @@ export class JupyterApplication implements IApplication, IDisposable {
     this._serverFactory.createFreeServer().catch(error => {
       console.error('Failed to create free server', error);
     });
+    installGlobalNavigationGuard();
```
krassowski, line 283: Should it happen earlier in the constructor?

krassowski (approved):
Thanks. Minor suggestion on potentially moving `installGlobalNavigationGuard` up in constructor to avoid surprises in the future if something gets added earlier.

If the goal is hardening, should we also handle `will-frame-navigate` (in a separate PR)?

PR #1094, opened 2026-07-28
Fix the Linux packages failing to launch, and verify each installer in CI

### LINKED ISSUE #870: Crash after upgrading to Ubuntu 24.04
## Description

```
jlab .
[514889:0921/151946.918521:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly.
Rather than run without sandboxing I'm aborting now.
You need to make sure that /opt/JupyterLab/chrome-sandbox is owned by root and has mode 4755.
Trace/breakpoint trap (core dumped)
```

## Reproduce

1. Install JupyterLab Desktop on Ubuntu 23.10
2. It works
3. Update to Ubuntu 24.04
4. It does not work

## Expected behavior

Works

## Context

- Operating System and version: Ubuntu 24.04
- JupyterLab-Desktop version: 4.2.5
- Installer: <s>snap</s> actually not sure

notluquis:
Closes #870. Follow-up to a review comment on #1091.

Three artifacts leave the linux leg and none was ever run. Each now gets launched in CI. The snap passed; the other two failed from the same cause: naming a value in electron-builder's config replaces its default instead of adding to it.

| | Symptom on first run | Cause |
| --- | --- | --- |
| deb | `zygote_host_impl_linux.cc:128] No usable sandbox!` on ubuntu-24.04, clean install (#870, open since 2024) | `deb.fpm`'s `--after-install` is appended after the built-in one and fpm keeps the last, so `after-install.tpl` never ran. The shipped postinst is 104 bytes and has no AppArmor block, though the profile itself is in `md5sums` |
| rpm | `error while loading shared libraries: libnspr4.so` | `rpm.depends: ["libXScrnSaver"]`, added in 2017, replaced all eight entries of `getDefaultDepends('rpm')`. REQUIRENAME in the released header has no `nss`, which is where libnspr4 comes from |

The fix vendors app-builder-lib 26.15.7's `after-install.tpl` plus our two jlab lines, wired through the documented `afterInstall` option instead of the raw fpm argument, and drops the rpm `depends` block. A unit test reads the upstream template out of `node_modules` and fails if the shipped copy drifts from it, so an electron-builder bump that touches the postinst breaks on the bump.

One loose end I did not take on here: nothing owns `/usr/bin/jlab`. The deb postinst creates it, the snap `configure` hook creates it, and the snap `remove` hook deletes it unconditionally, so with both installed the last writer wins. Giving it an owner needs the after-remove template vendored too.

The matrix also gets `fail-fast: false` and one retry on the installer build. Those are about this job rather than about the packages, so happy to split them out if you would rather review them separately.

One thing `fail-fast: false` changes that is worth deciding rather than inheriting: the snap publish step lives inside the matrix and gates only on `release-exists`, while `upload-release-assets` has `needs: publish`. So on a release run where one leg fails and linux succeeds, the snap reaches the store and the GitHub assets job is skipped. That race existed before, since cancellation is not instant, but this makes it systematic. Three ways out, and the choice is a release-pipeline call rather than a packaging one: leave it, move the snap publish into a job that needs the whole matrix, or drop `fail-fast: false` and accept losing three artifacts to one flake.

Worked through this with Claude Code; the postinst, postrm and rpm requires are read out of the released v4.6.2-1 packages, and the two failures plus the green run after them are this branch's own CI.


notluquis:
Green on the full matrix now, [run 30766360293](https://github.com/jupyterlab/jupyterlab-desktop/actions/runs/30766360293). Three commits, three verdicts, all from this branch's own CI:

| Commit | Linux leg | What it showed |
| --- | --- | --- |
| `e720238` | fail | deb: `No usable sandbox!`, exit 133. rpm never reached |
| `ab1a55a` | pass | deb and rpm both launch and print `4.6.2-1` |
| `331305f` | pass | reinstall and purge added: profile present after install, absent after purge |

The purge step also printed the one thing left open, which is why it is a note rather than an assertion:

```
note: /usr/bin/jlab is left behind by purge
```

Ready for review. Worked through with Claude Code; every line above is copied from those three runs.

notluquis:
Ready for review.

Since the last look, four gaps in the checks themselves: `/usr/bin/jlab` is pinned to its target rather than only tested for existence, the rpm step waits for Xvfb's socket instead of racing it, `snap wait system seed.loaded` runs best-effort first, and the drift assertion prints what diverged.

Copilot reviewed the current head, no comments.

One call left in the body: `fail-fast: false` lets a release run publish the snap while the assets job is skipped.

krassowski (approved):
Thank you!

Issue #786, opened 2024-03-02
electron-28 fails to start jupyterlab-desktop


PF4Public:
Starting `jupyterlab-desktop` with electron-28 fails with the following error:
```sh
16:37:22.419 › TypeError: Cannot read properties of undefined (reading 'filename')
    at Object.isDevMode (***/jupyterlab-desktop/build/out/main/utils.js:43:25)
    at ***/jupyterlab-desktop/build/out/main/main.js:192:55
    at new Promise (<anonymous>)
    at processArgs (***/jupyterlab-desktop/build/out/main/main.js:191:12)
    at App.<anonymous> (***/jupyterlab-desktop/build/out/main/main.js:169:15)
    at App.emit (node:events:517:28)
16:37:22.423 › null
16:37:22.424 › TypeError: Cannot read properties of undefined (reading 'filename')
    at Object.isDevMode (***/jupyterlab-desktop/build/out/main/utils.js:43:25)
    at getLogLevel (***/jupyterlab-desktop/build/out/main/main.js:55:17)
    at redirectConsoleToLog (***/jupyterlab-desktop/build/out/main/main.js:71:22)
    at App.<anonymous> (***/jupyterlab-desktop/build/out/main/main.js:178:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

This does not happen if starting with electron-27. I've hardcoded `isDevMode()` to always return `false`, which solved this issue for me.

I could not find the exact braking change which might've occured between those electron/node versions, it could very well be something on my system. Would you be able to verify whether you can also reproduce this issue?

JasonWeill:
@PF4Public Where are you seeing this issue? Are you trying to build JupyterLab Desktop locally or are you running a version that you downloaded and installed as a release? Thanks!

PF4Public:
I'm building it locally. It builds, but does not start with electron-28.

mbektas:
@PF4Public is there any particular reason you are trying to upgrade to electron 28? we are currently at electron 27.

mbektas:
It is probably failing due to [ESM module support ](https://www.electronjs.org/blog/electron-28-0#highlights)added to electron 28.

PF4Public:
> @PF4Public is there any particular reason you are trying to upgrade to electron 28? we are currently at electron 27.

[...]

PR #1097, opened 2026-07-29
Ask Electron whether the app is packaged instead of reading an undocumented internal

notluquis:
Refs #786.

```ts
export function isDevMode(): boolean {
  return require.main.filename.indexOf('app.asar') === -1;
}
```

becomes `!app.isPackaged`.

This is fragility rather than a live crash. Probing the Electron this repo ships, 42.5.0, launching a directory app the way `yarn start` does: master's expression returns `true` and `require.main.filename` is the literal string `electron`, with no throw. The upstream cause, electron/electron#40501, was closed in December 2023. That probe covers macOS, the bundled binary and the default-app path, not a system Electron on another platform.

What remains is that `require.main.filename` holding the string `electron` is an undocumented internal which already broke once, in the ESM work that produced #786. `app.isPackaged` is the documented API for the question `getAppDir` is actually asking. Behaviour for the shipped app is unchanged: `asar` is never disabled in the build config, and `app.isPackaged` is `basename(process.execPath) !== "electron"`, which agrees for every artifact we publish.

One thing worth deciding rather than assuming: `isDevMode()` answers three different questions and only two of them are the same.

| Call site | What it is really asking |
| --- | --- |
| `main.ts:240`, `process.argv.slice(isDevMode() ? 2 : 1)` | was this launched through Electron's default app |
| `utils.ts:36`, `getAppDir()` | is this an asar layout |
| `main.ts:81`, `:104`, log level | is this a packaged build |

Electron documents `process.defaultApp` for the first, and is explicit that it differs from `isPackaged`:

> When the app is started by being passed as parameter to the default Electron executable, this property is `true` ... even if the app is packaged (`isPackaged`) is `true`. This can be useful to determine how many arguments will need to be sliced off from `process.argv`.

So the argv slice arguably wants `process.defaultApp`. Not doing it here, since it changes argument parsing and deserves its own diff, but raising it so it is a decision rather than an oversight.

The darwin branch of `getJlabCLICommandTargetPath` carried a comment saying it could not be tested because it needed `require.main`. It is testable now, so it is tested. Mutation check: dropping the `!` fails all three new cases. Not covered: the packaged branch of `getAppDir`, where removing the `path.dirname` call fails nothing because the stub returns a path with no `.asar` in it. A case for that is a reasonable review request.

Worked through this with Claude Code; the probe against 42.5.0 and the mutation run are mine.


src/main/utils.ts
```diff
@@ -28,7 +28,9 @@ export interface ISaveOptions {
 }
 
 export function isDevMode(): boolean {
-  return require.main.filename.indexOf('app.asar') === -1;
+  // require.main is undefined under ESM and when the app runs against an
```
krassowski, line 31: In the future let's try to reign Claude from adding comments about deleted code

krassowski (approved):
Thank you!

Issue #824, opened 2024-05-07
Unexpected token '


OmniJax:
Upon opening JupyterLab I get an error message that says the following...
"A JavaScript error occurred in the main process

Uncaught Exception:
SyntaxError: Unexpected token '

![image](https://github.com/jupyterlab/jupyterlab-desktop/assets/74285021/b481569e-10c2-4fda-aac5-93f8aed39ffe)

I tried updating to the latest version but it didn't work

OmniJax:
When I try to create a new session, it gets stuck on this screen

![image](https://github.com/jupyterlab/jupyterlab-desktop/assets/74285021/895721a6-2434-41da-9766-c452c95b9b46)

OmniJax:
When I try to create a new session, it gets stuck on this screen

![image](https://github.com/jupyterlab/jupyterlab-desktop/assets/74285021/311c490d-616e-4a4b-b657-1456152e1a4f)

OmniJax:
> @OmniJax Thanks for opening this issue! From your first screenshot, I see the error is; `SyntaxError: Unexpected token ", ""… is not valid JSON`. How did you install JupyterLab Desktop? Which OS (name and version) are you using? If you launch JupyterLab Desktop from a terminal, do you see any errors in the terminal?

Hi! I am using win10, and I installed it by "JupyterLab-Setup-Windows-x64.exe".

Now I can use Jupyter-desktop after uninstalling the jupyterlab-desktop and deleting the folder 'C:\Users\XXX\AppData\Roaming\jupyterlab-desktop'. 

Now my problem is, when I try to create new session in welcome page, it will get stuck in “creating new session” stage. I can firstly execute “jupyter lab —no-browser” in cmd, and then connect to it in Jupyter-desktop, but it will be more troublesome. Actually I didn't have this problem before, it will help me automatically create a session and enter it.

thank you!

OmniJax:
@JasonWeill 

Hi! I am using win10, and I installed it by "JupyterLab-Setup-Windows-x64.exe".

Now I can use Jupyter-desktop after uninstalling the jupyterlab-desktop and deleting the folder 'C:\Users\XXX\AppData\Roaming\jupyterlab-desktop'. 

[...]

Issue #431, opened 2022-03-27
An error pops up: `A javaScript error occured in the main process`


hellojinwoo:
## Description
- An error message pops up after I execute the jupyterlab on desktop screen
```
A javaScript error occured in the main process

uncaught Exception:
Error: connect ECONNERFUSED 127.0.0.1:64279
   at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1142:16)
```
![image](https://user-images.githubusercontent.com/34431729/160287652-54eb7b53-a986-4e6d-9854-af12d0e44dc0.png)
- I have installed `Jupyterlab-desktop`, and 2 jupyter extensions, which are `darcula theme`(https://github.com/telamonian/theme-darcula) and `github extension`(https://github.com/telamonian/theme-darcula)
- 

## Expected behavior
- Can execute the jupyterlab-desktop with the error pop-up screen

## Context

- Operating System and version:  Windows 11
- JupyterLab-Desktop version: v.3.3.2-1
- Installer: .exe

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the terminal inside of JupyterLab-Desktop here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Browser Output (required for interface issues after loading)</summary>

<pre>
Paste the output from the DevTools JavaScript console here, if applicable.

</pre>
</details>

<details><summary>Logs (required for issues with loading, server and startup)</summary>

<pre>
Paste the logs from the `main.log` file here (optionally also from `renderer.log`).

</pre>
</details>

hellojinwoo:
- It seems to have a problem with `Server Connection Error`
![image](https://user-images.githubusercontent.com/34431729/160351934-d976cf49-d4fa-4f4d-b30e-78a897eaa27f.png)
- After I had the error pop-up screen `javaScript error occured in the main process`, my notebook gets disconnected from Python and it sets back to `no kernel`

hellojinwoo:
@mbektas Thank you for the feedback. Yes, I have constantly received the error message whenever I execute the jupyterlab. To solve the problem, I have deleted the jupyterlab-desktop as instructed in the installation page. And then I have reinstalled jupyterlab desktop. But the error keeps showing up.

Now it seems like the only solution would be formatting my laptop, which I wouldn't like to do for real. 

- Is there any other solution than formatting my laptop?

mbektas:
[...]

PR #1101, opened 2026-08-02
Warn when a config file could not be read, and ignore values of the wrong type

notluquis:
> **On hold, and back in draft.** This and #1115 both carry the atomic write, and this branch has the older copy. Rather than resolve that twice it will be rebuilt on top of #1115 once that lands, keeping only what is its own. Not worth reviewing in this shape.

## References

- Second half of #824 and #431; #1115 is the first half and this sits on top of it
- #1116 is the open question below

## Code changes

Values inside a file that parses are checked before they are used. `{"sessions": [null]}` threw on `key in null`, an unparseable date threw from `toISOString()` at save time, and a number reached `path.dirname`. Each is now measured against its declared default rather than a hand-written schema, so the check follows the declaration instead of drifting from it.

`persistSessionData` is honoured only as a real boolean, because it decides whether a remote server's cookies land on disk. Absent means true, since `serialize` omits the key in that case.

One dialog names the files that could not be read, with **Show in Folder** and **Continue**, and says nothing was moved. It waits for the first window and attaches to it: [Electron's docs](https://www.electronjs.org/docs/latest/api/dialog#dialogshowmessageboxwindow-options) note that on macOS a parentless message box runs synchronously, which held up the whole launch until somebody clicked.

## Open question, worth deciding before this merges

**#1116.** `save()` rebuilds the file from what survived `read()`, so rejecting a value deletes the key rather than ignoring it. Measured: `{"condaChannels":"conda-forge"}` comes back as `{}`. That is the argument that took an enum check out of this branch already, and it applies equally to `_hasShapeOf` and `pythonEnvsIn`, which are what is left. Keeping one and removing the other is not a line I can defend.

## User-facing changes

A message box at startup naming any configuration file that could not be read, instead of silently starting on defaults.

None beyond #1115's.

## Manual testing

macOS, corrupt and out-of-shape config through the real Electron runtime. The two ownership tests were repaired after measuring that removing the root-only guard left the whole suite green; it now turns one test red, and removing the chown turns two.

Not done: the startup notice has no e2e coverage. `stubAllDialogs` records nothing, so asserting it needs a recording wrapper, and the stub is installed after the first window while the notice fires as soon as that window is live.

## AI usage


notluquis:
Merge order note, since this one carries a conflict the others do not.

I trial-merged the eight open non-draft PRs onto `master` in sequence. Seven combinations are clean. The two that are not are this PR against #1077 and against #1087, both in `test/unit/utils.test.ts`, and both because all three append a new `describe` block to the end of that file. Nothing semantic: the resolution is to keep both blocks.

| Merged after | Result |
| --- | --- |
| #1097 then #1077 | clean |
| #1097 then #1087 | clean |
| #1101 then #1077 | conflict, `test/unit/utils.test.ts` |
| #1101 then #1087 | conflict, `test/unit/utils.test.ts` |

Cheapest order is therefore #1094, #1097, #1092, #1077, #1087, then this one last, which leaves one branch to rebase instead of two. I will rebase it as soon as whichever of those lands, so nothing here needs holding on my account.

Trial merges run locally with Claude Code; the table is what git actually reported, not a guess from the file lists.

notluquis:
Ready for review.

The one worth naming: a `settings.json` saved by Notepad or `Out-File` starts with a BOM, `JSON.parse` rejects it, and this branch was renaming that file away. It used to crash loudly; here it became a silent revert of `pythonPath` and everything else. Verified with the real bytes.

Three smaller ones, each a case where the notice went missing: an array parsed as usable, a failed rename told nobody, and the notice ran after the constructor inside the same try.

Copilot flagged that the dialog said "settings" when the list can hold `app-data.json`; it now says configuration.

krassowski (commented):
There are some robustness issues highlighted in a local Opus review, I looked at the first one but did not verify all; the overall choices made so far look reasonable.

> High:
> - When the rename fails, the app overwrites the corrupt file with defaults and destroys the data the dialog says is safe. The catch at utils.ts:124-128 leaves the file in place and records the live path, so the dialog reads "Nothing was deleted. Each is still on disk, here: .../settings.json". Then will-quit at app.ts:637-640 runs userSettings.save(), which does fs.writeFileSync(userSettingsPath, ...) at settings.ts:234 and writes {}, because every setting sits at its default after the failed read.
>    - The same write fires within seconds of any settings change, so the user does not even need to quit.
> - A second app instance renames the config aside and can never say so. The quarantine happens during module import, before app.requestSingleInstanceLock() is called at all, and then handleMultipleAppInstances() rejects at main.ts:330, the catch runs app.quit(), and reportUnreadableConfig() on the following line never executes. Double-clicking a notebook while the app is running is enough: the notebook opens in the first instance and the user's edited settings.json is silently moved.
> 
> Medium:
> - A symlinked config file is unlinked rather than quarantined. fs.renameSync operates on the final path component without following it, so a settings.json symlinked into a dotfiles repo becomes settings.json.corrupt pointing at an untouched target, and the next save() creates a fresh regular file in its place. From then on every edit the user makes in their dotfiles repo has no effect and nothing explains why. Verified by running the rename against a symlink.
> 
> Low:
> - The new corrupt-file test in appdata.test.ts:62 leaves existsSync returning true for every path and readFileSync returning corrupt bytes, and the only beforeEach in that block calls resetAppData(), which touches no mock. It passes because every following test happens to reassign both stubs. This PR added exactly the guard that prevents this to settings.test.ts:37-45, with a comment naming the hazard, and skipped this file.
>

notluquis:
Split. The atomic write and the read guard are now #1115; this keeps the value checking, the notice and the CLI reporting, and sits on top of it. The branch here still contains #1115, so the diff shows both halves until that lands, at which point a merge from master reduces it to the second.

The reason is worth writing down rather than just doing. This had reached sixteen files and about two thousand added lines across two concerns that do not need each other. Three review passes each turned up something severe, and the pattern in the third was that fixes written to close the previous round's findings were producing the next round's worst ones: an exemption meant to stop a mid-session hand edit from disabling saves let a later reader write `{}` over a project's settings, and an enum membership check meant to catch typos deleted `uiMode: custom` from machines that had it, because `save()` rebuilds the file from whatever survived the read. Both replaced a crash with silent data loss, and both are reverted here.

One thing found on the way that was not about this code at all: the `fs` mock in `settings.test.ts` and `workspacesettings.test.ts` returned `undefined` from `readFileSync`, which threw a `TypeError` with no `code` at module import and marked the settings path before any test ran. Both files then spent their whole run against a config the production code considered unreadable, and the flagship #824 regression test was passing on that rather than on its own fixture. It was also a live flake, since the file only passed in declaration order. Fixed in 07077f6, and the mutation check now lands where it should.

Nothing here is waiting on review feedback; posting so the split is on the record rather than a surprise.

Issue #1049, opened 2026-07-15
tracking issue: CI and release pipeline hygiene


notluquis:
Tracking the CI and release-pipeline hygiene work. These are separate from the dependency modernization (that's its own tracker), and none of them are user-facing. Grouping so the pieces don't drift apart.

### Sub-issues
- publish workflow hygiene (dedup, concurrency, caching, hardening) #1045
- macOS installer builds every arch in each job #1044

### Related PR
- [ ] zizmor workflow hardening #1017

### Net-new (sub-issue below)
A fast lint gate on PRs. Right now eslint and prettier only run inside the heavy publish job (which needs conda and a four-platform installer build), so lint feedback is slow and coupled to the release path.

Drafted with Claude Code, checked against the workflows at 200a939.

PR #1112, opened 2026-08-17
Run the unit suite on macOS and Windows too

notluquis:
## References

- Part of #1049
- The `.gitattributes` here would also have prevented the CRLF postinst behind #870

## Code changes

The unit job ran on `ubuntu-latest` only. It now runs on macOS and Windows too. The first Windows run had thirteen tests red, which is why the fixes are in this PR and not a follow-up: a matrix nobody can merge is worth nothing.

Twelve are tests, failing the same way. The suite stubs `process.platform` to reach a branch, but `path` binds `win32` or `posix` when it is imported, from the real host. So the function takes the branch the test asked for and joins the result the host's way, and `/env/bin/python` meets `\env\bin\python`. The file already knew, next to one of them:

```ts
// path.join uses host OS separator — just verify it doesn't include Scripts
```

Where the assertion is about layout it now folds separators before comparing (`toSlash` is the identity on the two legs that were already green). Where it is about which program gets spawned, the platform is pinned instead. Two were also weak and got exact: `envPathForPythonPath` was checked with `toContain('/env')`, true of its own argument.

The thirteenth is real. `electron-builder-scripts/linux_after_install.sh` becomes the `postinst` of the deb and rpm, and the Windows runner checked it out with CRLF. A CRLF shebang fails with `bad interpreter: /bin/bash^M: no such file or directory`; with `.gitattributes` in place the same checkout writes LF, so the attribute is what does it. Five files pinned. No tracked file has CRLF today, so nothing is renormalised.

Coverage thresholds stay enforced once, on ubuntu.

## User-facing changes

None.

None. `publish` is `needs: test`, so a Windows-only failure now blocks the installers and therefore a release, where before only ubuntu could. That is the point, and it is a new way to be held up.

## Manual testing

The Windows leg is the testing: thirteen red to green across two runs on this PR. Legs cost ubuntu 25s, macOS 41s, Windows 1m05s against a 10 minute cap.

Locally on macOS: full suite, `tsc --noEmit`, eslint, prettier. The two strengthened assertions were mutation-checked, since those could have been weakened rather than fixed.

Not done: nothing here runs the app on Windows. The e2e matrix is still ubuntu and macOS, which is #1113.

## AI usage

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# These run as Unix shell scripts wherever they end up, so they keep LF even
+# when checked out on Windows. linux_after_install.sh becomes the postinst of
+# the deb and rpm, where a CRLF shebang is read as an interpreter named
+# "/bin/bash\r" and the install fails.
+*.sh text eol=lf
+electron-builder-scripts/snap-hooks/* text eol=lf
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,8 +14,16 @@ permissions:
 
 jobs:
   test:
-    name: 'Unit tests'
-    runs-on: ubuntu-latest
+    # One leg per OS, the way jupyterlab splits linuxtests / macostests /
+    # windowstests. The suite mocks electron, so no binary is needed and the
+    # cost is a runner each; what it buys is the filesystem behaviour that only
+    # differs by platform, which the config tests reach directly.
+    name: 'Unit tests (${{ matrix.os }})'
+    strategy:
+      fail-fast: false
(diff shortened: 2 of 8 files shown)


krassowski:
> test/unit/config-fs.test.ts

Is this a local uncommitted file?

krassowski (approved):
Thanks!

notluquis:
It comes from #1115, which adds it, so it is not in this branch or on master yet. The sentence in #1115 that mentions it points here for the matrix, which is probably how it landed in this tab.

Issue #1116, opened 2026-08-17
A config value the reader rejects is deleted from the file at the next save


notluquis:
`UserSettings.save` and `ApplicationData.save` rebuild their file from memory. `read` skips a value it does not like, so that value never reaches memory, and the next save writes the file back without it.

Rejecting is not ignoring. It is deleting, at the next save, silently.

Measured on `fix/config-json-guard`, driving the real `UserSettings` against a real file:

```
before: {"condaChannels":"conda-forge","theme":"dark"}
after a read and a save: {"theme":"dark"}
```

`condaChannels` is declared as an array of strings and holds a bare string, which is the shape #1101's guard exists to keep out of `join()`. The guard works. The cost is the user's value gone from disk, with a line in a log nobody opens and no dialog, because the file parsed and so nothing marked it.

## The CLI is an active producer of that value

This is the part that separates it from a historical-data problem. `jlab config set condaChannels conda-forge` runs `JSON.parse('conda-forge')`, which throws, and falls back to `JSON.parse('"conda-forge"')` — so it writes the **string**. `handleEnvSetCondaChannelsCommand` writes `argv._.slice(1)`, an **array**. Two commands of the same CLI write the same setting in incompatible shapes, and the reader added in #1101 rejects one of them.

So the value does not only arrive from old builds with no migration, which was the reason the enum membership check was taken back out of #1101. It arrives from the shipped CLI today. The two cases are not symmetric, and this one has a fix at the writing end: coerce the parsed value to the setting's declared shape at `config set`, or refuse it there with a message, rather than accepting it and dropping it a launch later. `serverEnvVars` takes a string the same way.

## What #1118 changes, and what it does not

#1118 makes `save` read the file and merge over it rather than rebuilding, for `settings.json` and `.jupyter/desktop-settings.json`. Where that lands, a rejected value is no longer deleted: it stays on disk untouched, which is the behaviour this issue asks for.

It does not close the whole thing:

- `ApplicationData.save` still rebuilds `app-data.json` from memory, and the loss reported against that file is dropped **array entries** rather than unknown top-level keys, which needs a different fix than merging.
- A value that stays on disk but is refused at every read is inert rather than lost, which is better, and still silent. Nothing tells the user their `condaChannels` line does nothing.
- The producer above is untouched either way.

Read this issue against `master`, where the rebuild is still what happens, and re-read it after #1118 lands.

## Where the shape appears

| where | what goes |
| --- | --- |
[...]

PR #1118, opened 2026-08-20
Keep the config keys this build has no setting for

notluquis:
## References

- Found while reviewing #1101, but older than it and independent of it: this reproduces on `master` with nothing else applied
- #1116 is the wider version of the same mechanism, for values the reader rejects rather than keys it does not know
- Worth landing before #1101, which makes rejection safe there for free

## Ordering against the other config pull requests

Measured with `git merge-tree --write-tree --name-only`, not assumed, and confirmed with a real `git merge --no-commit` that was then aborted. The three-argument `git merge-tree <base> <a> <b>` reports no conflicts for the first row, so it is the wrong form to ask:

| | |
| --- | --- |
| this x #1115 | **conflicts**, in `src/main/config/settings.ts` and `test/unit/settings.test.ts` |
| this x #1101 | same conflict, since that branch contains #1115 |
| this x #1112 | clean |
| each against `master` | clean |

`read()` and `save()` are what both this and #1115 rewrite, so whichever lands second needs a rebase. **#1115 first is the cheaper order**: the local `readJsonFileOrEmpty` here exists only because the shared reader lands there, so rebasing onto it is deleting this helper and calling `readJsonConfigFile`, plus `save` returning the writer's boolean. The reverse order works too and is no worse in substance, just more to redo.

#1112 is independent of all of these and can land whenever.

## What this does not fix

A file that is there and could not be read is now left alone rather than merged over. That started as #1115's job and turned out to be this branch's: the reason for deferring it was that an `EBUSY` or `EACCES` would fail the write too, and it does not. Measured on darwin, a `settings.json` at mode `0200` gives `EACCES` on the read and `OK` on the write, because `writeFileSync` opens `O_WRONLY` and never reads, so the file ended as `{}` with one log line. The same holds for the hand-edit with a trailing comma that `troubleshoot.md` tells people to make.

`save()` returns a boolean so a caller can tell, and `jlab config set` and `jlab config unset` both ask before printing success. Counted on the current head: four of the twenty-three call sites use the result and nineteen ignore it; #1115 changes the same signature and is where the rest get wired.

With the app open, `jlab config set logLevel debug` in a separate process is still reverted at quit: `will-quit` calls `userSettings.save()`, which writes every `SettingType` key from in-memory state that predates the other process's write. Unknown keys now survive that; known ones do not. Same as `master`, so not a regression, and not something this claims.

## Code changes

`save` rebuilds `settings.json` and `desktop-settings.json` from the values it holds, and `read` only ever looked at the keys in `SettingType`. A key the file had and this build has no setting for was not ignored. It was deleted at the next save.

Measured on `master`, nothing else applied:

```
before: {"futureSetting":42,"theme":"dark"}
after a read and a save: {"theme":"dark"}
```

So a `settings.json` written by a newer build loses whatever an older one does not recognise, and `troubleshoot.md` sends readers to edit this file by hand, so a line somebody added on purpose goes the same way.

### What the references do

| | on write | can it lose a key |
| --- | --- | --- |
| VS Code | splices edits into the original text by offset, `applyEdits` in `vs/base/common/jsonEdit.ts` | no, nothing outside the edit is touched |
| `conf`, which `electron-store` wraps | reads the whole file into an object, changes one key, writes the object back: `const {store} = this` through `this.store = store` | no, unknown keys survive by construction |
| here, before this | rebuilds from `key in SettingType` | yes, silently |

This takes conf's shape: it re-reads the file inside the write and merges there. An earlier commit on this branch cached at read time and replayed the snapshot at save, which is not the same thing and is where its defects came from; that is reworked.

VS Code's approach keeps comments and key order too, but it means a JSON-with-offsets editor, which is a larger
change than the problem needs.

### The change

`save` merges over what the file holds rather than rebuilding it, so what this build has no setting for is kept and written back. `read` is unchanged: it still walks the enum.

`read` walks `SettingType` rather than the file, which is what `master` did and is why `master` was immune to a `__proto__` key: nothing out of the file ever indexes `_settings`.

`save` decides three things per key, not two. A key of `SettingType` matching its default is deleted, because it does not belong in the file whatever the file says. A key a project cannot override is left alone. Everything else is written. Collapsing the last two into one boolean is what let a stale on-disk value outlive the change that replaced it, and the commits on this branch have that mistake in them.

A key that is not overridable by a project is kept too. It does nothing in a project file, but deleting a line somebody wrote is worse than leaving one that has no effect.

## User-facing changes

A key in `settings.json` or `.jupyter/desktop-settings.json` that this build does not recognise stays there. `jlab config list` does not show it, because it is not a setting; nothing else changes.

None. Every key that used to round-trip still does; the ones that used to vanish do not.

## Manual testing

The before-and-after above is a run against `master` rather than a description of one, driving the real `UserSettings` against a real file.

Five new branches, mutation-checked one at a time, each turning exactly one test red: either `save` dropping the leftovers, either `read` not collecting them, and the workspace `unsetValue` not clearing them.

The three cases above were written as failing tests before the fix, and two of them turned out to be testing something else: one passed because its fixture had no `serverArgs`, so it asserted on the key's absence rather than its removal, and another drove `unsetValue` with `theme`, which no project can override and the CLI refuses. Both are fixed; that is the only reason the shape came out right this time.

Not done: none of the [Review guidance](https://github.com/jupyterlab/jupyterlab-desktop/blob/master/dev.md#review-guidance) checks were run on Windows or Linux.

## Remaining

`ApplicationData.save` rebuilds the same way. `app-data.json` is app-managed rather than documented for hand editing, and the loss reported against it is dropped array entries rather than unknown top-level keys, which needs a different fix. #1116.

## AI usage

### DIFF
--- a/src/main/cli.ts
+++ b/src/main/cli.ts
@@ -1028,6 +1028,7 @@ function handleConfigSetCommand(argv: any) {
     return;
   }
 
+  let saved: boolean;
   if (projectPath) {
     const setting = userSettings.settings[key];
     if (!setting.wsOverridable) {
@@ -1037,10 +1038,18 @@ function handleConfigSetCommand(argv: any) {
 
     const wsSettings = new WorkspaceSettings(projectPath);
     wsSettings.setValue(key as SettingType, value);
-    wsSettings.save();
+    saved = wsSettings.save();
   } else {
     userSettings.setValue(key as SettingType, value);
-    userSettings.save();
+    saved = userSettings.save();
+  }
+
+  // save now declines when the file is there and could not be read, so printing success without asking would tell somebody their setting landed while the file was left untouched
+  if (!saved) {
+    console.error(
+      `Could not write the settings file, so "${key}" was not saved. The log names the file and why.`
+    );
+    return;
   }
 
   console.log(
--- a/src/main/cli.ts
+++ b/src/main/cli.ts
@@ -982,7 +982,7 @@ function handleConfigListCommand(argv: any) {
   console.log(listLines.join('\n'));
 }
 
-function handleConfigSetCommand(argv: any) {
+export function handleConfigSetCommand(argv: any) {
   const parseSetting = (): { key: string; value: string } => {
     if (argv._.length !== 3) {
       console.error(`Invalid setting. Use "set <settingKey> <value>" format.`);
@@ -1059,7 +1059,7 @@ function handleConfigSetCommand(argv: any) {
   );
 }
 
-function handleConfigUnsetCommand(argv: any) {
+export function handleConfigUnsetCommand(argv: any) {
   const parseKey = (): string => {
     if (argv._.length !== 2) {
       console.error(`Invalid setting. Use "unset <settingKey>" format.`);
@@ -1082,6 +1082,7 @@ function handleConfigUnsetCommand(argv: any) {
     return;
   }
 
+  let saved: boolean;
   if (projectPath) {
     const setting = userSettings.settings[key];
     if (!setting.wsOverridable) {
@@ -1091,10 +1092,18 @@ function handleConfigUnsetCommand(argv: any) {
 
     const wsSettings = new WorkspaceSettings(projectPath);
     wsSettings.unsetValue(key as SettingType);
-    wsSettings.save();
+    saved = wsSettings.save();
   } else {
     userSettings.unsetValue(key as SettingType);
-    userSettings.save();
+    saved = userSettings.save();
+  }
+
+  // the same refusal `set` already reports: a reset announced over a file that was left untouched is the one shape automation cannot tell from success, and the guard was written on one side only
+  if (!saved) {
+    console.error(
+      `Could not write the settings file, so "${key}" was not reset. The log names the file and why.`
+    );
+    return;
   }
 
   console.log(
--- a/src/main/config/settings.ts
+++ b/src/main/config/settings.ts
@@ -213,17 +213,19 @@ export class UserSettings {
     const data = fs.readFileSync(userSettingsPath);
     const jsonData = JSON.parse(data.toString());
 
-    for (let key in SettingType) {
-      if (key in jsonData) {
-        const setting = this._settings[key];
-        setting.value = jsonData[key];
+    this._unclaimed = {};
+    for (const key of Object.keys(jsonData)) {
+      if (key in SettingType) {
+        this._settings[key].value = jsonData[key];
+      } else {
+        this._unclaimed[key] = jsonData[key];
       }
     }
   }
 
   save() {
     const userSettingsPath = UserSettings.getUserSettingsPath();
-    const userSettings: { [key: string]: any } = {};
+    const userSettings: { [key: string]: any } = { ...this._unclaimed };
 
     for (let key in SettingType) {
       const setting = this._settings[key];
@@ -241,6 +243,9 @@ export class UserSettings {
     );
   }
 
+  // what the file held and this build had no setting for. Kept because save
+  // rebuilds the file, so a key left out of it is a key deleted from disk.
+  protected _unclaimed: { [key: string]: any } = {};
   protected _settings: { [key: string]: Setting<any> };
 }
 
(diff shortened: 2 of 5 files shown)


notluquis:
Draft. This is wrong in a way a follow-up commit does not fix, and the approach has to change.

`read` walks the file and asks `key in SettingType` for each key. `in` walks the prototype chain, so `__proto__` answers true, `this._settings['__proto__']` resolves to `Object.prototype`, and the assignment that follows lands on it. That runs at module import, through the singleton at the bottom of `settings.ts`, and `troubleshoot.md` is the document that tells people to edit this file by hand.

Confirmed rather than reasoned about: a probe left such a file behind on my machine, and the next unrelated test run failed while registering its suites, before reaching any test of mine. Deleting the file restored the suite. Anything importing the module inherits it. `toString`, `constructor` and `hasOwnProperty` are misclassified by the same check and get dropped rather than preserved, so the goal of this change also fails for exactly those keys.

The enum-first loop this replaced was immune, because a name out of the file never indexed anything.

Adjacent, from the same review and each verified:

| | |
| --- | --- |
| a deleted `desktop-settings.json` is recreated on the next save, because the kept keys make the write guard true | the deletion used to stick |
| `setValue` does not clear the matching kept key, so for a non-overridable one the file keeps the old value while `getValue` returns the new one | in-memory and on-disk diverge and stay diverged |
| a project-overridable key whose value equals the global one is still dropped | same loss this exists to fix, so the claim in the description does not hold |
| the reset of the kept keys sits below the early return, so a re-read after the file disappears writes the previous contents back | `read` is not idempotent |

The description also mis-states the reference it cites. `conf` re-reads the file inside the write (`const {store} = this`, mutate, `this.store = store`); this caches at read time and replays the snapshot later. That difference is what produces all four rows above. Doing what conf actually does, re-reading and merging inside `save`, removes the two extra fields, the base-and-subclass split, and the invariant that currently rests on a comment and one test. It also needs an own-property check regardless, which is what closes the first item.

Not pasting the payload. `Object.hasOwn` in place of `in`, and null-prototype objects for the bags, are the mechanism.

Reworking it as read-modify-write rather than patching these one at a time, and not tonight: this is the fourth time today that a fix of mine in this area produced something worse than what it replaced, and that pattern is the reason to stop rather than push on.

Worked through this with Claude Code, which also produced the review that caught it.

notluquis:
Reworked rather than patched, since the findings all came from one choice.

The first attempt cached what the file held at read time and replayed it at save. Doing the merge in `save` instead removes the reason `read` had to walk the file at all, so it goes back to walking `SettingType`, which is what master did and is why master was immune: nothing out of the file ever indexes `_settings`.

| | before | now |
| --- | --- | --- |
| `__proto__` in the file | reached `Object.prototype` at module import | stays an own property and round-trips |
| a deleted `desktop-settings.json` | recreated on the next save | stays deleted |
| `setValue` on a non-overridable key | file kept the old value, `getValue` returned the new one | file gets the new one |
| an overridable key equal to the global value | dropped | kept |
| a file whose top level is an array | rewritten as `{"0":1,"1":2}` | yields nothing |
| a key this build does not know | kept | kept |

Both extra fields are gone, and with them the one-set-per-class invariant that was defended by a comment and a test. Deleting a key now needs an explicit `unsetValue`, tracked in a set: a value equal to its default cannot be told apart from one nobody set, so `save` cannot use that to decide a removal without deleting what somebody typed.

The description also cited `conf` and did not match it. It re-reads inside the write; that is what this does now, and the body has been corrected.

Four mutations, one red test each: `read` walking the file, `save` not merging, the unset set ignored, the reader accepting non-objects. The fourth passed on the first attempt because prettier had reformatted the ternary and the mutation never applied. Worth stating plainly, because it is the same shape as the finding itself: the test was fine and my check was the thing that could not fail.

Separately, the `fs` mock in `settings.test.ts` had no `readFileSync` and no `beforeEach`, so the new tests would have leaked their stubs into whatever ran after them. It throws an ENOENT-coded error now and resets per test, verified across four shuffled runs.

Worked through this with Claude Code. Every row in that table is a run against the real classes and a real file.

Issue #261, opened 2021-09-28
Windows Defender detects virus


Carreau:
Hi, 

This was reported on the security mailing list, but I don't believe there is any actual vulnerability, it's more a concern. 

It looks like windows defender is finding that the installer has a virus here are the screenshots provided by original reporter. I'll send them here.
![d2](https://user-images.githubusercontent.com/335567/135127781-00d8072c-da72-4cbf-a03c-6a01514e14d5.PNG)

![d1](https://user-images.githubusercontent.com/335567/135127809-ff6767e1-9a8b-47e8-b661-79da798346b7.PNG)

mbektas:
@Carreau please refer to this document
https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-smartscreen/microsoft-defender-smartscreen-overview

Our app is not code-signed yet. It could be possible reason.

"Microsoft Defender SmartScreen determines whether a downloaded app or app installer is potentially malicious by:

Checking downloaded files against a list of reported malicious software sites and programs known to be unsafe. If it finds a match, Microsoft Defender SmartScreen shows a warning to let the user know that the site might be malicious.

**_Checking downloaded files against a list of files that are well known and downloaded by many Windows users. If the file isn't on that list, Microsoft Defender SmartScreen shows a warning, advising caution_.**"

alexander-manley:
This is how SUSE Rancher Kubernetes Desktop has solved the electron Microsoft code-signing for their opensource project:
https://github.com/rancher-sandbox/rancher-desktop/pull/405

PR #1127, opened 2026-08-26
Fix the macOS installers failing to sign, and the cross-architecture dmg that hung

notluquis:
## References

- Blocks the v4.6.3-1 release: every installer leg failed or hung, so `upload-release-assets` was skipped and the pre-release carried no assets at all.
- Upstream: [electron-builder#10066](https://github.com/electron-userland/electron-builder/issues/10066), open since 7 August, no comments and no fix.
- Related: #1018, which moved the x64 leg to a native Intel runner because `conda pack` aborts under Rosetta.

## Code changes

Three separate faults, each measured on CI before and after.

**Signing.** `security set-key-partition-list` failed with `SecKeychainUnlock: the passphrase is not correct` on the x64 leg, deterministically and through the built-in retry, while arm64 signed the same secrets in the same run. electron-builder creates its own temporary keychain with a random password and then unlocks it with the `.p12` import password instead. That is harmless while macOS lets the call through on an already-unlocked keychain, and macOS 26.6 stopped letting it through: `macos-26-intel` runs 26.6, `macos-26-arm64` runs 26.5.2. Reproduced on macOS 27, where the same call with the correct password reaches `SecItemCopyMatching` and with the wrong one gives exactly the error CI reports.

Dropping `CSC_LINK` and `CSC_KEY_PASSWORD` from the installer step makes electron-builder skip that path and sign through `CSC_IDENTITY_AUTO_DISCOVERY`, which finds the identity in the `build.keychain` the Codesign Application Server step already imported into. #10066 calls this workaround unusable in CI for want of a keychain to discover from; this workflow builds one twenty lines earlier. Both variables stay on that step, which needs them to decode and import the certificate. The keychain therefore has to outlive `dist:osx`, so the `security delete-keychain` that ran at the end of the signing step is gone rather than moved: every runner here is GitHub-hosted, the machine is destroyed when the job ends, and GitHub's guide scopes that cleanup to self-hosted runners. `default-keychain -s` becomes `list-keychains -d user -s`, which is the pattern [GitHub's guide](https://docs.github.com/en/actions/use-cases-and-examples/deploying/installing-an-apple-certificate-on-macos-runners-for-xcode-development) uses: adding to the search list is all `codesign` and auto-discovery need, while making it the default and then deleting it left the session pointing at a keychain that no longer existed.

**The cross-architecture hang.** With signing fixed, both legs then sat until they were killed. The logs are symmetric: each one packages, signs, zips and dmgs its own architecture and completes, then starts the other and goes silent, 19 minutes on Intel and 30 on Apple Silicon. It is the cross-architecture dmg specifically, twice out of twice in opposite directions, rather than the intermittent `dmgbuild` the comment above already covers.

`mac.target` declared `arch` on both targets, and `targetFactory` only falls back to the command line when a target leaves it unset, so `--x64` and `--arm64` were ignored and each leg built both architectures. Removing it lets each build its own. Both `dmg` and `zip` survive, which the update feed needs.

**The runaway.** The publish job had no `timeout-minutes`, so it inherited GitHub's 360 and one hang cost six hours of a macOS runner, billed at ten times the Linux rate. The test job above already sets 10; this sets 90, against a measured 33 minutes for the macOS x64 leg, part of which is Apple's notary queue and has no bounded turnaround.

**Failing loudly.** Without `CSC_LINK`, an identity auto-discovery cannot find no longer stops the build: electron-builder throws only for MAS or `forceCodeSigning` and warns otherwise, and `platformPackager` then skips the `afterSign` hook, so `scripts/notarize.js` does not run either. A locked keychain or a future change to `find-identity` would have produced a green job publishing an unsigned, un-notarized dmg, where the same situation used to be a hard error. `forceCodeSigning: true` restores that, and does not reach `dist:osx-dev`, where `isSignAllowed()` returns false first.

## User-facing changes

None on macOS: the installers are signed with the same identity as before, `Developer ID Application: NumFOCUS, Inc.`, verified identical in the logs on both sides of the change.

On Windows the `.exe` is no longer signed, and that needs saying plainly even though nothing changes for a user. The `env:` block here is shared by all four matrix legs, and electron-builder's `getCscLink("WIN_CSC_LINK")` falls back to `CSC_LINK`, so removing it takes the Windows signature with it: four `certificateFile` lines in the previous run's log, none in this one.

What it was signing with was the Apple certificate. Read out of the published v4.6.2-1 `.exe`, the signer is `Developer ID Application: NumFOCUS, Inc.`, issued by Apple's Developer ID CA. It carries the Code Signing EKU so `signtool` accepts it, but the Microsoft included-CA report lists 549 roots and Apple is not among them, so the chain never validated on Windows: users saw the Unknown Publisher prompt either way, which is what #261 reports. Nothing configured this; `build.win` has no signing options and there is no `WIN_CSC_LINK`.

So this drops a signature that did not work rather than one that did, and the fix is a Windows certificate from a CA Microsoft trusts, which is #202. Detail and measurements are in a comment there. Restoring the fallback for the Windows leg alone is a one-line conditional if you would rather keep the state exactly as it was.

`yarn dist:osx` with no architecture flag now builds for the host rather than for both, which is what running it locally usually wants. `yarn dist:osx-64` and `yarn dist:osx-arm64` are unchanged in intent and now actually honour it. `dist:osx-dev` and `dist:osx-arm64-dev` are untouched.

## Why the checks on this pull request are red

Not the change. This is a pull request from a fork, so `secrets` arrive empty, and the release path runs anyway: `Find Release with tag` looks for a draft or pre-release matching `package.json`, finds the published v4.6.3-1, and the signing steps then fail on `security create-keychain -p ""` with exit 255 and on the snap login. Every fork pull request will do this while that pre-release stands, which is #1128; #1123 passed at 19:00 on 25 August only because the release was still a draft then, and `listReleases` does not return drafts to a read-only token.

The changes here were exercised on `release-v4.6.3-1`, where the secrets are present, and that is the run the table below reports.

## Manual testing

All four installer legs pass on `release-v4.6.3-1` for the first time, run 32912875766:

| | before | after |
| --- | --- | --- |
| macOS x64 | failed at signing, three times | passes |
| macOS arm64 | 25 min, then hung until killed | 13 min |
| Linux x64 | 15 min | 13 min |
| Windows x64 | 9 min | 8 min |

The keychain behaviour was reproduced on macOS 27 locally, and the architecture resolution checked against `targetFactory` directly rather than inferred. zizmor on this file goes from 32 findings to 30 and adds none. What is not covered here: launching the produced `.dmg`, which is the check #870 exists for and needs a person.

## AI usage


krassowski:
I deleted the pre-release and will close and reopen to test CI

package.json
```diff
@@ -131,25 +131,18 @@
...
-          ]
+          "target": "zip"
         }
       ],
       "entitlements": "build/entitlements.plist",
       "entitlementsInherit": "build/entitlements.inherit.plist",
       "darkModeSupport": true,
       "hardenedRuntime": true,
       "gatekeeperAssess": false,
+      "forceCodeSigning": true,
```
krassowski, line 145: This might turn out to be problematic on `master`, I will open a follow-up PR if it does.

krassowski (approved):
Thanks!

